Issue #26, opened 2015-03-25
Excessive wakeups


afcowie:
_powertop(8)_ on Linux reports the most expensive process in terms of processor wakups is _ghcid_.

```
     Usage     Events/s     Category       Description
  4.5 ms/s     108.8        Process        ghcid --command=cabal exec ghci tests/Experiment.hs
```
this represents about a quarter of the user-space power draw on this machine when on battery!

Originally I thought you were using _inotify(7)_ so assumed that it wouldn't be doing any wakeups at all. When I saw it sucking down power like this, I had a look and just realized you're **not** using [hnotify](http://hackage.haskell.org/package/hinotify). Oh.

Is that deliberate? Assuming not, would you countenance a dependency on [fsnotify](http://hackage.haskell.org/package/fsnotify) (which purports to be a cross-platform wrapper around the different filesystem notification mechanisms)?

AfC

ndmitchell:
I have issues with ghcid on battery power too. Depending on fsnotify is fine, but last time I tried on windows it failed. I'll try again tomorrow.

ndmitchell:
My reports of it not working are at https://github.com/haskell-fswatch/hfsnotify/issues/50. I suggest the right thing to do is add an abstraction layer in ghcid, and an option to use either notify or polling. We can give a flag for the alternatives, and if it seems not to work robustly, default to polling on Windows.

afcowie:
I'd agree with that. 'hnotify' exposes a lot more of inotify, and given how many corner cases there are  (c.f. http://blogs.operationaldynamics.com/andrew/software/haskell/rebuilding-via-inotify ) hiding that breadth of features behind a lowest common denominatior abstraction layer always worries me. For instance, inotify can watch files or a directory for a wide range of events whereas 'fsnotify' only talks about directories.

ndmitchell:
I was thinking of the abstraction layer being based on what ghcid needs, not what the notify thing provides. I want to wait for a list of files, so one API could be:

    waitCreate :: IO Waiter
    waitFor :: Waiter -> [FilePath] -> IO [FilePath]

Create a `Waiter` object, given that object, wait for some files and return which ones changed. That should map to all reasonable abstractions and make the most of them all.

ndmitchell:
All done, just broken, so dealing with that in the next ticket.

PR #27, opened 2015-04-07
Inotify

afcowie:
So I took a stab at using inotify to watch files. This is a bit of a mess; it took me a while to realize that the add watch function _wasn't_ blocking. So I forked it off on a Haskell thread, but this then raises the question of what happens to all the other watches; I don't know how strong `withINotify` is, but it probably needs to be managed by an Async, etc.

Anyway, this is obviously Linux specific but handles watching a list of files. If you were to refactor the code to want to watch a list of _directories_ then you could use **fsnotify** pretty much as-is (and it has a polling mode for platforms without inotify/kpoll/whaver). 

Anyway, a patch in the spirit of "here's something, I'm sure you can do better; either way all the best"

AfC

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -1,6 +1,5 @@
 :set -fwarn-unused-binds -fwarn-unused-imports
-:set -isrc -itest
-:load Ghcid src\Paths.hs Test
+:set -isrc -itest -idist/build/autogen
 :def docs_ const $ return $ unlines [":!cabal haddock"]
 :def docs const $ return $ unlines [":docs_",":!start dist\\doc\\html\\ghcid\\Language-Haskell-Ghcid.html"]
 :def test const $ return $ unlines ["Test.main"]
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 /.dist-buildwrapper
 /.project
 /cabal.sandbox.config
+/tags
+/.cabal-sandbox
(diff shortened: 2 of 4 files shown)


afcowie:
Oh; ignore what I said about directories (and using fsnotify for that matter); you've gone to all the trouble of finding out what sources _ghci_ actually loaded, so might as well only watch those.

[In the past I've watched entire source trees and rebuild on any change therein, which does work; absent NFS it's not like anything else is changing things in there. If you decide to adopt that instead then yes we can watch nominated directories]

I'll see what I can do about rewriting this patch to make it more sensible if you don't get there first.

AfC

ndmitchell:
I got there first - see Wait.hs in the source tree. In about two hours I should have cleaned it up enough to actually work.

ndmitchell:
All done, and turned on by default. Seems to work for me. Can you check out the HEAD and give it a try?

ndmitchell:
Wow, it feels much snappier using the notify based code. I guess instead of waiting a whole 0.1s it kicks into action immediately, which is noticeable on my fast desktop, where sleep really is the dominating factor.

ndmitchell:
(Closing the pull request, but I'd still welcome feedback in #26)

afcowie:
I was hacking on something else today, but I'll have a chance to poke
around in the morning. Cheers.

ndmitchell:
I'm hoping you will just need a bit of real world usage, and no hacking time at all, for me I'm using it just as normal.

afcowie:
@ndmitchell power usage is indeed no longer an issue. Well done. There's a different problem arising from this implementation, though; I'll raise a new issue.

PR #30, opened 2015-04-21
First attempt at testing automatically after a successful reload

cschneid:
This is rough and probably shouldn't yet be merged.

Use with something like: `ghcid --command="cabal repl ghcid_fasttest" --testcommand="main"`

Wishlist / more features:

- [ ] Render test output nicer
- [ ] Interrupt (?) test output on file save
- [ ] Stream test output
- [ ] Interpret exit code perhaps? Fold a successful test run back into "All good"


ndmitchell:
Thanks for the code. I merged in your diff, and checked a load of it in as separate pieces and with modifications. I haven't done any of your next steps, but it does work as it stands.

PR #37, opened 2015-06-11
Adds option to disable shell title update

nkpart:
I'm running ghcid within a terminal in Emacs and updating the shell title isn't supported, so it just spews the characters out.

Added a flag to disable the shell title update

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -35,6 +35,7 @@ data Options = Options
     ,height :: Maybe Int
     ,width :: Maybe Int
     ,topmost :: Bool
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks! All merged, and I'll be making a release in about half an hour with this change.

I did a little refactoring, turning the setTitle into just a boolean. Better to keep the complexity down where possible. The only reason setOutput is a function is so that the test suite can capture it, but I don't yet have automated tests for the title, so no need to go that far (yet...).

nkpart:
Ah yep, that makes sense, thanks :)

Issue #35, opened 2015-05-07
Change the titlebar icon on Windows


ndmitchell:
Could have Red for failure, Green for success, Yellow for warnings.

ndmitchell:
Fixed in #40 by @WJWH (many thanks!)

PR #40, opened 2015-06-15
Change icons depending on number of errors

WJWH:
Fix for issue #35. Has some minor issues like not changing the icon back if you kill ghcid with ctrl-c, but nothing major that I could find. 

I added a conditional dependency on Win32 for windows users and kept all code within CPP blocks, so it should only affect windows users. I tested it on a linux box and it still builds fine there.

### DIFF
--- a/ghcid.cabal
+++ b/ghcid.cabal
@@ -33,8 +33,9 @@ library
         extra >= 1.2,
         process >= 1.1,
         cmdargs >= 0.10,
-        terminal-size >= 0.3,
-        Win32
+        terminal-size >= 0.3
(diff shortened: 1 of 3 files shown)


ndmitchell:
Awesome, thanks! I was originally thinking of constructing bitmaps etc for the icons, but reusing the message box icons is significantly nicer. I've modified the code a bit, refactored some pieces etc. The only real additional functionality is that I capture the icons before and restore them after, so it restores the icon if you do a Ctrl-C.

In limited testing today I'm already finding it useful functionality, far more than I thought I would :)

PR #46, opened 2015-10-02
Whitespace and stack.yaml file

meiersi:
I usually drop trailing whitespace on save, which would mess up with quite a few files in this repo. So I opted for introducing this commit before I start working on interruptible test suites.

I also like the ease of setting up compilation using stack, which is why I added the stack.yaml file.

Please feel free to ignore any of these commits, if they don't suit your taste.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@
 /tags
 /.cabal-sandbox
 /.stack-work
+# Use 'stack init' on a fresh checkout to generate your own stack.yaml file
+# for easy building.
+/stack.yaml
--- a/ghcid.cabal
+++ b/ghcid.cabal
@@ -24,8 +24,8 @@ source-repository head
 
 library
   hs-source-dirs:  src
-  default-language: Haskell2010  
-  build-depends:   
+  default-language: Haskell2010
+  build-depends:
         base >= 4,
         filepath,
         time,
@@ -36,7 +36,7 @@ library
         terminal-size >= 0.3
   if os(windows)
     build-depends: Win32
-  other-modules:   
+  other-modules:
                    Paths_ghcid,
                    Language.Haskell.Ghcid.Types,
                    Language.Haskell.Ghcid.Parser,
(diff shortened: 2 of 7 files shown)


ndmitchell:
I dislike trailing whitespace, but don't configure my editors that way because so many other people have it in code that I edit. But a diff dropping all trailing whitespace is most gratefully accepted.

For the stack file, what's the purpose of adding it? Can't you just run `stack init` very quickly each time you checkout `ghcid`? I don't like the idea of adding extra metadata such as the lts version I depend on since it's not checked by my buildbot and may go stale. I am gradually switching to stack, so I'm not against it if there is a benefit beyond saving `stack init`.

meiersi:
The stack file is currently just used for convenience. I'll drop the commit
from the PR once I'm back home on my machine.

Neil Mitchell <notifications@github.com> schrieb am Fr., 2. Okt. 2015 11:30
AM:

meiersi:
...and regarding doing `stack init` on every checkout. Would it then be OK
to add the stack.yaml to the .gitignore together with a comment about why
it is being ignored?

Simon Meier <simon.y.meier@gmail.com> schrieb am Fr., 2. Okt. 2015 4:18 PM:

ndmitchell:
If it provides genuine convenience, say beyond typing `stack init`, because it needs some manual tweak or similar, then I'm happy to include it.

ndmitchell:
Certainly, in the .gitignore makes a lot of sense - I think I already ignore the .stack-work.

meiersi:
Yes, you do ignore .stack-work. I've added the corresponding .gitignore change, which means this request would be ready to merge.

ndmitchell:
Thanks!

yairchu:
> If it provides genuine convenience, say beyond typing `stack init`, because it needs some manual tweak or similar, then I'm happy to include it.

I think it a `stack.yaml` adds the convenience of assured reproducible build and alleviated doubts.
Without one, any update of a dependency package may break the build, making it a mystery of which dependency versions the code was built and tested with..

ndmitchell:
@yairchu Agreed, I think there is a lot of value in knowing exactly which versions of dependencies work together - which is why I keep `ghcid` in Stackage, so you can grab it from there and be sure everything works nicely.

Looking back to the patch from @meiersi, I now do have my editor configured to drop trailing spaces, and I also documented my `stack.yaml` preferences at https://github.com/ndmitchell/neil#quirks.

Issue #10, opened 2014-10-09
Accept key input


ndmitchell:
@fegu suggests that instead of `All good` the tool should say something like:

```
All good
   Press R to reload
   Press P to pause
   Press B to build
   Press H to HLint
```

I am particularly keen on the R option, which would kill ghci and restart from scratch, as it provides some workaround for #7, especially if you are using `cabal repl` and change the `.cabal` file. The P option would also be useful when I'm on a train, at the moment `ghcid` sucks my battery flat in about the duration of my train ride, sometimes slightly sooner.

ndmitchell:
Of the options above, R to reload is mostly unnecessary with `--reload` and `--restart`, P is mostly unnecessary now we have proper watching of files, B is mostly unnecessary with the interruptible `--test` support - so I'm left wondering if there is anything good to hang off the keyboard. Things like `hlint` or `docs` are probably best done from a real ghci session.

PR #51, opened 2016-03-02
Interruptible test command

luigy:
First pass at stealing ordeal's approach
Wanted to get this rolling so hopefully someone picks it up :)
 here is a 30 second [clip](https://cldup.com/tu8euZobN1.mp4) of the new behavior

* added a `--spawn` flag that would background the test command and restart it on file change.

TODO:

- [x] control-c should kill the --test command before exiting. Currently the spawned process is left running.
- ~~[ ] stream tests ?~~
- [ ] Interpret exit code perhaps? Fold a successful test run back into "All good"
- ~~[ ] Render test output nice~~

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -19,6 +19,10 @@ import System.Directory.Extra
 import System.Exit
 import System.FilePath
 import System.IO
+#if !defined(mingw32_HOST_OS)
+import System.Posix.Signals (signalProcess, sigINT)
+import System.Process.Internals (ProcessHandle,ProcessHandle__(..), withProcessHandle)
+#endif
 
 import Paths_ghcid
 import Language.Haskell.Ghcid
@@ -34,6 +38,7 @@ data Options = Options
     {command :: String
     ,arguments :: [String]
     ,test :: Maybe String
+    ,spawn :: Bool
     ,height :: Maybe Int
     ,width :: Maybe Int
     ,topmost :: Bool
@@ -49,6 +54,7 @@ options = cmdArgsMode $ Options
     {command = "" &= typ "COMMAND" &= help "Command to run (defaults to ghci or cabal repl)"
     ,arguments = [] &= args &= typ "MODULE"
     ,test = Nothing &= name "T" &= typ "EXPR" &= help "Command to run after successful loading"
+    ,spawn = False &= name "S" &= help "Spawn the test command to run after successful loading"
     ,height = Nothing &= help "Number of lines to use (defaults to console height)"
     ,width = Nothing &= help "Number of columns to use (defaults to console width)"
     ,topmost = False &= name "t" &= help "Set window topmost (Windows only)"
@@ -126,7 +132,7 @@ main = withWindowIcon $ ctrlC $ do
                 return (f width 80 (pred . fst), f height 8 snd)
         withWaiterNotify $ \waiter ->
             handle (\(UnexpectedExit cmd _) -> putStrLn $ "Command \"" ++ cmd ++ "\" exited unexpectedly") $
-                runGhcid waiter (nubOrd restart) command outputfile test height (not notitle) $ \xs -> do
+                runGhcid waiter (nubOrd restart) command outputfile test spawn height (not notitle) $ \xs -> do
                     outWith $ forM_ (groupOn fst xs) $ \x@((s,_):_) -> do
                         when (s == Bold) $ setSGR [SetConsoleIntensity BoldIntensity]
                         putStr $ concatMap ((:) '\n' . snd) x
@@ -137,8 +143,8 @@ main = withWindowIcon $ ctrlC $ do
 data Style = Plain | Bold deriving Eq
 
 
-runGhcid :: Waiter -> [FilePath] -> String -> [FilePath] -> Maybe String -> IO (Int,Int) -> Bool -> ([(Style,String)] -> IO ()) -> IO ()
-runGhcid waiter restart command outputfiles test size titles output = do
+runGhcid :: Waiter -> [FilePath] -> String -> [FilePath] -> Maybe String -> Bool -> IO (Int,Int) -> Bool -> ([(Style,String)] -> IO ()) -> IO ()
+runGhcid waiter restart command outputfiles test spawn size titles output = do
     let outputFill :: Maybe (Int, [Load]) -> [String] -> IO ()
         outputFill load msg = do
             (width, height) <- size
@@ -151,7 +157,7 @@ runGhcid waiter restart command outputfiles test size titles output = do
     restartTimes <- mapM getModTime restart
     outStrLn $ "Loading " ++ command ++ " ..."
     nextWait <- waitFiles waiter
-    (ghci,messages) <- startGhci command Nothing True
+    (ghci, ph, messages) <- startGhci command Nothing True
     curdir <- getCurrentDirectory
 
     -- fire, given a waiter, the messages, and the warnings from last time
@@ -195,13 +201,22 @@ runGhcid waiter restart command outputfiles test size titles output = do
             outputFill (Just (loadedCount, messages)) ["Running test..." | isJust test]
             forM_ outputfiles $ \file ->
                 writeFile file $ unlines $ map snd $ prettyOutput loadedCount $ filter isMessage messages
-            whenJust test $ \test -> do
-                res <- exec ghci test
-                outputFill (Just (loadedCount, messages)) $ fromMaybe res $ stripSuffix ["*** Exception: ExitSuccess"] res
-                updateTitle ""
+
+            let outputTest res = do
+                    outputFill (Just (loadedCount, messages)) $ fromMaybe res $ stripSuffix ["*** Exception: ExitSuccess"] res
+                    updateTitle ""
+            testCmd <- case test of
+                Just test | spawn ->
+                    onceFork $ exec ghci test
+                Just test ->
+                    return $ exec ghci test
+                _ -> return $ return []
+            when (isJust test && not spawn) $
+                outputTest =<< testCmd
 
             when (null wait) $ do
                 putStrLn $ "No files loaded, nothing to wait for. Fix the last error and restart."
+                stopGhciTest ph test spawn
                 exitFailure
             reason <- nextWait $ restart ++ wait
             outputFill Nothing $ "Reloading..." : map ("  " ++) reason
@@ -209,11 +224,14 @@ runGhcid waiter restart command outputfiles test size titles output = do
             if restartTimes == restartTimes2 then do
                 nextWait <- waitFiles waiter
                 let warnings = [m | m@Message{..} <- messages, loadSeverity == Warning]
+                stopGhciTest ph test spawn
+                when (isJust test && spawn) $ outputTest =<< testCmd
                 messages <- reload ghci
                 fire nextWait messages $ Just warnings
             else do
+                stopGhciTest ph test spawn
                 stopGhci ghci
-                runGhcid waiter restart command outputfiles test size titles output
+                runGhcid waiter restart command outputfiles test spawn size titles output
 
     fire nextWait messages Nothing
 
@@ -238,3 +256,21 @@ prettyOutput loaded [] = [(Plain,allGoodMessage ++ " (" ++ show loaded ++ " modu
 prettyOutput loaded xs = concat $ msg1:msgs
     where (err, warn) = partition ((==) Error . loadSeverity) xs
           msg1:msgs = map (map (Bold,) . loadMessage) err ++ map (map (Plain,) . loadMessage) warn
+
+
+stopGhciTest :: ProcessHandle -> Maybe String -> Bool -> IO ()
+#if defined(mingw32_HOST_OS)
+stopGhciTest _ _ _ = return ()
+#else
+stopGhciTest ph test spawn = do
+  when (isJust test && spawn) $ controlC ph
+  where
+    controlC ph = do
+        mPid <- getPid ph
+        whenJust mPid $ signalProcess sigINT
+
+    getPid ph = withProcessHandle ph (return . go)
+      where
+        go (OpenHandle x)   = Just x
+        go (ClosedHandle _) = Nothing
+#endif
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -153,7 +153,7 @@ runGhcid waiter restart command outputfiles test spawn size titles output = do
     restartTimes <- mapM getModTime restart
     outStrLn $ "Loading " ++ command ++ " ..."
     nextWait <- waitFiles waiter
-    (ghci, ph, messages) <- startGhci command Nothing True
+    (ghci, messages) <- startGhci command Nothing True
     curdir <- getCurrentDirectory
 
     -- fire, given a waiter, the messages, and the warnings from last time
@@ -212,7 +212,7 @@ runGhcid waiter restart command outputfiles test spawn size titles output = do
 
             when (null wait) $ do
                 putStrLn $ "No files loaded, nothing to wait for. Fix the last error and restart."
-                interrupt ph test spawn
+                interrupt ghci test spawn
                 exitFailure
             reason <- nextWait $ restart ++ wait
             outputFill Nothing $ "Reloading..." : map ("  " ++) reason
@@ -220,12 +220,12 @@ runGhcid waiter restart command outputfiles test spawn size titles output = do
             if restartTimes == restartTimes2 then do
                 nextWait <- waitFiles waiter
                 let warnings = [m | m@Message{..} <- messages, loadSeverity == Warning]
-                interrupt ph test spawn
+                interrupt ghci test spawn
                 when (isJust test && spawn) $ outputTest =<< testCmd
                 messages <- reload ghci
                 fire nextWait messages $ Just warnings
             else do
-                interrupt ph test spawn
+                interrupt ghci test spawn
                 stopGhci ghci
                 runGhcid waiter restart command outputfiles test spawn size titles output
 
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -19,10 +19,6 @@ import System.Directory.Extra
 import System.Exit
 import System.FilePath
 import System.IO
-#if !defined(mingw32_HOST_OS)
-import System.Posix.Signals (signalProcess, sigINT)
-import System.Process.Internals (ProcessHandle,ProcessHandle__(..), withProcessHandle)
-#endif
 
 import Paths_ghcid
 import Language.Haskell.Ghcid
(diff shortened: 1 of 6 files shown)


ndmitchell:
Thanks! I was just thinking of making a release of ghcid, and bumping this feature to the next release, but given your progress I'll revise my plans and hope to get this incorporated. Looking at the code so far, two things spring to mind:

* I'd push the `ProcessHandle` type inside `Ghci` as a field, and have `stopGhciTest` (perhaps named `interrupt`?) be a member of the API in `Language.Haskell.Ghcid`. The initial applicability test bit would need pulling out into the main loop.
* Why would anyone _not_ want to run with `--spawn`? Are there any downsides?

luigy:
>  I was just thinking of making a release of ghcid, and bumping this feature to the next release, but given your progress I'll revise my plans and hope to get this incorporated.

Hehe yes I noticed lots of changes came in :)

> I'd push the ProcessHandle type inside Ghci as a field, and have stopGhciTest (perhaps named interrupt?) be a member of the AP

:+1: yes, interrupt is better

> Why would anyone not want to run with --spawn? Are there any downsides?

I think I remember you mentioning performance I somewhere in the issue so I thought maybe you didn't want it as the default, but I probably I interpreted it wrong.

I just realized that I've forgotten to handle outputting test output when the background process is done since I was trying it only for non terminating commands

ndmitchell:
In fact, I think I'll release now anyway, since I've got it in a nice state, and there are a lot of improvements. Having two releases in quick succession is no bad thing anyway, and letting this change bake a bit isn't a bad idea either.

My performance concerns were all with spawning a separate process to run the test, which will be a lot slower. If we can interrupt using the existing `ghci` process then it seems like purely a win. We might want a streaming version of `exec` at some point, to make it easier to relay test output as it comes back.

luigy:
> In fact, I think I'll release now anyway, since I've got it in a nice state, and there are a lot of improvements. 

I agree :)
more releases with new features are always well received 

> My performance concerns were all with spawning a separate process to run the test, which will be a lot slower. If we can interrupt using the existing ghci process then it seems like purely a win.

I see

> We might want a streaming version of exec at some point, to make it easier to relay test output as it comes back.

:+1: yes

--

A bit late to the train, but ghcid is awesome and fast for dev!! :zap: :zap:

ndmitchell:
> A bit late to the train, but ghcid is awesome and fast for dev!! :zap: :zap:

One reason for wanting to get a new release out is that today I doubled the speed for some of my projects - Shake in GHCid goes from 7s to 3s :)

luigy:
> - Shake in GHCid goes from 7s to 3s :)

:zap: :fire: @ndmitchell :fire: :zap:

reiddraper:
This is looking great! Thanks @luigy

ndmitchell:
@luigy - do you want to take another crack at this pull request taking into account my comments, or would you prefer me to merge your changes and then make my suggestions on top? I'm happy either way.

luigy:
@ndmitchell I've update the PR and I think this should now work on windows as well :raised_hands: 
Also, took a stab at handling ctrl-c. Let me know if anything can be improved :)

> I'd push the ProcessHandle type inside Ghci as a field

done :+1:. Is that what you meant?

> The initial applicability test bit would need pulling out into the main loop

I'm not sure what you mean here or also how to properly test this :joy: 

>Why would anyone not want to run with --spawn? 

If the approach seems reasonable it should be ready for cleanup and probably taking a stab at making spawn the default

src/Language/Haskell/Ghcid/Types.hs
```diff
@@ -8,9 +8,10 @@ module Language.Haskell.Ghcid.Types(
 
 import Data.Typeable
 import Control.Exception.Base (Exception)
+import System.Process (ProcessHandle)
 
 -- | A GHCi session. Created with 'startGhci'.
-newtype Ghci = Ghci (String -> IO [String])
+newtype Ghci = Ghci (ProcessHandle, String -> IO [String])
```
ndmitchell: I'd just flip this to a real data type, `data Ghci = Ghci ProcessHandle (String -> IO [String])`

src/Language/Haskell/Ghcid.hs
```diff
@@ -94,4 +105,9 @@ stopGhci ghci = handle (\UnexpectedExit{} -> return ()) $ void $ exec ghci ":qui
 
 -- | Send a command, get lines of result
 exec :: Ghci -> String -> IO [String]
-exec (Ghci x) = x
+exec (Ghci (_,f)) = f
+
+-- | Interrupt the test command.
+interrupt :: Ghci -> Maybe String -> Bool -> IO ()
+interrupt (Ghci (ph,_)) test spawn =
+    when (isJust test && spawn) $ interruptProcessGroupOf ph
```
ndmitchell: The API should be `interrupt :: Ghci -> IO ()` and the checks should be in the non-API half.

src/Language/Haskell/Ghcid/Types.hs
luigy: derp! of course :)

luigy:
@ndmitchell I was playing with this https://github.com/luigy/ghcid/compare/interruptible-test-command...luigy:interruptible-test-command-rfc
thoughts?

src/Language/Haskell/Ghcid.hs
```diff
@@ -77,8 +83,13 @@ startGhci cmd directory echo = do
                     Just msg -> return msg
     r <- parseLoad <$> f ""
     writeIORef echo False
-    return (Ghci f,r)
 
+    let ghci = Ghci (ph, f)
+#ifndef mingw32_HOST_OS
+    tid <- myThreadId
+    installHandler keyboardSignal (Catch (interrupt ghci (Just "") True >> stopGhci ghci >> throwTo tid ExitSuccess)) Nothing
+#endif
```
ndmitchell: What is the installHandler doing? As far as I can tell, if you press a key in the GUI, it interrupts the running ghci process then kills the entire process. Is this an orthogonal feature? e.g. Ticket #10? Why is it calling interrupt, given exiting the parent process kills the child process anyway? Why both interrupt and stopGhci?

ndmitchell:
For info, I'm happy to clean up those little pieces after merge, other than the interrupt handler, which I don't understand enough yet.

The RFC stuff all seems like sensible direction to go, from that I can see:

* You invert the Ghci type so it contains the methods you can run on it, rather than the data it needs to run. Sensible, although then it probably wants moving out of Types, and also using field names rather than positional data types.
* You add locking around the operations, which seems like a good idea. I would need to read the locking carefully, but very sensible plan.
* You are adding an interactive test output process. I'd probably call it something else (not `withTestResult`, but `execWithOutput` or similar - not sure of a good name). But it looks like a good addition. Any chance `exec` can be implemented on top of the more general version? Fewer primitives is always good.

src/Language/Haskell/Ghcid.hs
luigy: >What is the installHandler doing?

This takes care of ctrl-c not terminating my spawned test process(which can run forever) when exiting ghcid.  If there is another way, let me know :)

>As far as I can tell, if you press a key in the GUI, it interrupts the running ghci process then kills the entire process. Is this an orthogonal feature?

`keyboardSignal` is actually an alias of `sigINT`.  I know! misleading name, right, but probably has a historical meaning? :joy: 

> Why is it calling interrupt, given exiting the parent process kills the child process anyway? Why both interrupt and stopGhci?

This should be only interrupt + trowTo. 
`stopGhci` is a leftover from different approaches I was trying. By the way, stopGhci without trowTo left the process hanging

luigy:
here is another short [clip](https://cldup.com/zU-xtcjhZW.mp4) showing off rfc branch

>You invert the Ghci type so it contains the methods you can run on it, rather than the data it needs to run.  

I wasn't sure if we really wanted to expose process handle? (are there any downsides if we do?)

>Sensible, although then it probably wants moving out of Types, and also using field names rather than positional data types.

Yes, it should just be record, right?  I was going for something similar to `exec` in Ghci and ended up with my rfc branch

> You add locking around the operations, which seems like a good idea. I would need to read the locking carefully, but very sensible plan.

the `testLock` is to guarantee that `interrupt` or `execTest` are not happening at the same time to avoid a possible false positive of `testRunning` Var.  Also calling execTest simultaneously would discard all the following test commands until the first one has finalized or been interrupted.

>You are adding an interactive test output process. I'd probably call it something else (not withTestResult, but execWithOutput or similar - not sure of a good name). But it looks like a good addition. 

`execWithOuput` :+1: I like this

>Any chance exec can be implemented on top of the more general version? Fewer primitives is always good.

I'll try something out

mgsloan:
On unix systems, I recommend using `pid <- forkProcess myMain`, and later doing `signalProcess sigKILL pid`.  See my comment [here](https://github.com/haskell/haskell-ide-engine/issues/2#issuecomment-150425675).  This is helpful in circumstances where terminating the main thread does not cause a full termination of the application.

ndmitchell:
No need to expose all the Ghci fields, just some - exposing the process handle seems like a bad idea, since it leaks implementation detail.

For the interrupt thing, I still don't fully understand. Are you saying your GHCi test command spawns a process itself? This is Linux, so if you kill the parent process, shouldn't everything die anyway? I don't understand the reason for doing anything special on Ctrl-C.

reiddraper:
> On unix systems, I recommend using pid <- forkProcess myMain, and later doing signalProcess sigKILL pid. See my comment here. This is helpful in circumstances where terminating the main thread does not cause a full termination of the application.

@mgsloan doesn't this defeat the whole purpose of this being fast? If you kill the entire ghci process, you have to restart it. I would say if interrupting your function doesn't clean up threads it spawns, then, Don't Do That.

ndmitchell:
Before figuring out how to send interrupts and deal with signals I'd like to be convinced that we need to go anywhere near this area at all - as far as I can tell sending Ctrl-C kills ghcid which should end up with ghci dying which should end up with spawned tasks dying. Windows might need more (explicitly kill ghci), but this code is non-Windows only.

mgsloan:
@reiddraper `forkProcess` creates a new process separate from your Ghci process, which efficiently uses a copy-on-write scheme in memory.  So, you get an entirely new process which can be killed without terminating your ghci session.  This works extremely well in practice, I have written a ghci driver that uses this approach.  ide-backend also used this approach...

reiddraper:
@mgsloan huh, ok. I would have been worried about:

> forkProcess comes with a giant warning: since any other running threads are not copied into the child process, it's easy to go wrong: e.g. by accessing some shared resource that was held by another thread in the parent.

But I guess in this specific use-case it might be safe.

mgsloan:
Yes, I don't recall any problems with using forkProcess for this.  ide-backend tries to control stdio pretty tightly, and there's some trickiness there, but if you don't care about that stuff, it's really straight forward.

One example of a shared resource is, say, an MVar.  You can't use MVars for inter-process communication - mutating it in one process will not mutate it in the other.

luigy:
>For the interrupt thing, I still don't fully understand. Are you saying your GHCi test command spawns a p rocess itself? 

> Before figuring out how to send interrupts and deal with signals I'd like to be convinced that we need to go anywhere near this area at all - as far as I can tell sending Ctrl-C kills ghcid which should end up with ghci dying which should end up with spawned tasks dying. 

I meant that after Control-C `ghciddd` process died, but  the spawned "ghci"(and it's child processes)  from `createProcess` did not.

Here are some images of the behavior I'm seeing without handling Control-C...

## create_group=False
### Running
running
### Control-C on ghciddd
Control-C

## create_group=True
### Running
running
### Control-C on ghciddd
Control-C



> This is helpful in circumstances where terminating the main thread does not cause a full termination of the application.

:hand: Wait a second! are you saying that ghciddd's child processes like `ghci`, `cabal ghci`, etc(created by `createProcess`) might not always receive a signal on `Control-C`?
if so, this explains why my spawned process(from createProcess) was left alive when hitting Control-C on ghciddd

>On unix systems, I recommend using pid <- forkProcess myMain, and later doing signalProcess sigKILL pid. See my comment here.

what do we mean by later doing? catching UserInterrupt?
...
> . This works extremely well in practice, I have written a ghci driver that uses this approach. ide-backend also used this approach.

...wait, what!!!  so you mean it as an alternative to sending sigINT for _reloading_ (:reload) + _rerunning_ (:test_cmd) the "test" command passed to ghciddd? :open_mouth:

>forkProcess creates a new process separate from your Ghci process, which efficiently uses a copy-on-write scheme in memory. So, you get an entirely new process which can be killed without terminating your ghci session.

!!!!! I have to go see some code + test it + inspect it with htop to let this sink in :)

I guess this could be an alternative to the current approach, but what about the Control-C issue?

What an odd behavior on Control-C with create_group=False :confused: 
@ndmitchell I think that at least in `unix` it is probably more reliable to use a combination of `createProcess` with `create_group=True` and `installHandler`(which uses sigaction under the hood) to get more guaranteed control what actually happens on Control-C.  
From what I understand the create_group gives us a fresh group process id that we can send signals with more confidence.
I'm sure y'all been probably doing this a bit longer than me and I want to learn, so let's discuss :)

-- 

@ndmitchell also apart from the Control-C issue does the reloading of test command work of windows?
I was hoping maybe 2c966aa was enough :cold_sweat:

mgsloan:
￼> Wait a second! are you saying that ghciddd's child processes like ghci, cabal ghci, etc(created by createProcess) might not always receive a signal on Control-C?
if so, this explains why my spawned process(from createProcess) was left alive when hitting Control-C on ghciddd

Yes, when the child ghci process is running some code, it might not handle SIGINT properly if it gets caught by the running code.  Both ghcid and the child process will receive SIGINT, but if the child doesn't handle it properly, it will continue running.

> > On unix systems, I recommend using pid <- forkProcess myMain, and later doing signalProcess
> > sigKILL pid. See my comment here.
>
> what do we mean by later doing? catching UserInterrupt?

Ahh, I misunderstood the featureset / scope of ghcid.  I thought it did something like `yesod devel`, where on code change, it reloads, kills the old application, and runs the new one.  So, off the topic of this PR for running tests.

> !!!!! I have to go see some code + test it + inspect it with htop to let this sink in :)

You can read about it here! https://en.wikipedia.org/wiki/Fork_(system_call)

> I guess this could be an alternative to the current approach, but what about the Control-C issue?

I'm not sure I grok all the details of the Ctrl-C issue, but this forkProcess stuff may be helpful.  At least that way, you can see if it's the child GHCI process, or its child, the forked process interpreting the tests, which is causing issues.

luigy:
>Ahh, I misunderstood the featureset / scope of ghcid. I thought it did something like yesod devel, where on code change, it reloads, kills the old application, and runs the new one. So, off the topic of this PR for running tests.

the "test" name is misleading imo, but at least I was using it like yesod devel in the original [clip]( https://cldup.com/tu8euZobN1.mp4) right? not sure if the original scope was only test, though :)

the flow is something like... ghciddd with "test" command=":main port" will start my server
then wait for file changes of loaded files...
on file change -> send interrupt to "ghci" -> :reload -> rerun "test" command (":main port")

luigy:
>This stuff is particularly helpful for web servers. If you kill the main thread of the web server but other threads remain running, it can hold on to the port binding. Then, the next time the application is launched, it can't get the port, and you're stuck running the old version of the application. With forkProcess, you can force things to get cleaned up.

looks like what I'm trying to achieve with using fresh process group?

>It might be cool to do something even fancier. It'd be cool if there was a way to ask the RTS about what threads are running. Then it might be possible to do the kill on the main thread, and wait a bit for the app to tear down all its threads. If it doesn't do so promptly, detect this case, warn the developer about it, and kill the process.

mgsloan:
> on file change -> send interrupt to "ghci" -> :reload -> rerun "test" command (":main port")

I'm not sure how important this is, but another benefit of using `forkProcess` is that you can wait till `:reload` has finished before killing the old one.

> looks like we are really trying not to handle Control-C directly? is there a particular reason? I mean I'm not trying to do anything fancy with it, but only enforcing that every child process is completely killed

It seems fine to me to try to enforce this - I'm ambivalent.  I just figured I'd bring up the `forkProcess` stuff because it seems quite relevant and I found it made my internal yesod-devel-alike more reliable when working on projects with bad async exception behavior.

luigy:
>I'm not sure how important this is, but another benefit of using forkProcess is that you can wait till :reload has finished before killing the old one.

hmm interesting
currently there is a lock that only let's one command at a time to "write" to `ghci` so no `:reload` can happen until the "test" command is done or interrupted.  
I'm probably missing something and I haven't looked at ghci code to achieve what you're saying :)

>It seems fine to me to try to enforce this - I'm ambivalent. I just figured I'd bring up the forkProcess stuff because it seems quite relevant and I found it made my internal yesod-devel-alike more reliable when working on projects with bad async exception behavior.

was this more in particular to the "reloading" behavior of it, rather than when you're done with yesod-devel-alike and just wanted to exit? or both?

mgsloan:
> currently there is a lock that only let's one command at a time to "write" to ghci so no :reload can happen until the "test" command is done or interrupted.

Hmm, good point.  In order to do this 100% right, some fancy stuff needs to be done with stdio for the forked process.  This wouldn't be able to use this for programs that expect stdin, because it would go to the ghci process which is still running a repl.

No modification of ghci itself should be necessary (in line with the goals of ghcid)

> was this more in particular to the "reloading" behavior of it, rather than when you're done with yesod-devel-alike and just wanted to exit? or both?

Mostly relevant to the reloading behavior.

ndmitchell:
I'm very keen to avoid forkProcess unless it produces significant benefits and can be done in a way that abstracts over the cross-platform differences between Windows and Linux. I develop on Windows and will continue to do so, so the less chunk of code I can't test the better.

I've got a paper deadline on Wednesday, but will take a look at reading all the comments, understanding the issues, and merging the code, after that. Thanks all for your thoughtful insights into the issue - Ctrl-C seems pretty subtle.

src/Ghcid.hs
```diff
@@ -195,15 +195,22 @@ runGhcid waiter restart command outputfiles test size titles output = do
...
+                whenLoud $ outStrLn $ "%TESTING: " ++ t
+                forkIO $ do
+                    res <- exec ghci t
+                    whenLoud $ outStrLn "%TESTING: Completed"
+                    outputFill (Just (loadedCount, messages)) $ fromMaybe res $ stripSuffix ["*** Exception: ExitSuccess"] res
+                    -- setSGR [Reset]
+                    updateTitle ""
+                return ()
 
             when (null wait) $ do
```
luigy: If we can move this check before executing test command we can be sure we can drop the nested interrupt

ndmitchell:
Thanks for the advice and code everyone. I've merged, and will be going through the code in detail tomorrow morning to clean it up, partly incorporating the changes from the rfc above, and partly just to make sure I understand all the details and try it out.

luigy:
:tada: :clap: :+1:

@ndmitchell I've noticed that `:reload` might be interrupted in some cases and rerun test command with old code

I had some local changes that I've forgotten to push here that only sends interrupt if a test command is running https://github.com/luigy/ghcid/commit/c1a4760b91c0989966ea27f20a152a0b9042f0ef

ndmitchell:
@luigy They look handy, so I've merged as #54 and will look at in the same go.

PR #67, opened 2016-04-23
--reload/--restart searches all non-dotted subdirs

simonmichael:
--reload/--restart DIR only watches source files within DIR. This change makes it also search subdirectories recursively, except for directories whose name begins with a dot.


ndmitchell:
Thanks! Nice idea.

PR #68, opened 2016-05-11
Suppress status messages flag

pkamenarsky:
I have autosave turned on in my editor. ghcid works great, however the constant flickering whenever a file is autosaved is really distracting, especially with many warnings filling the screen.

This option disables the `"Reloading..."` messages that clear the screen and cause the flickering.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 /.project
 /cabal.sandbox.config
 tags
-/tags
 /.cabal-sandbox
 /.stack-work
 # Use 'stack init' on a fresh checkout to generate your own stack.yaml file
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /.dist-buildwrapper
 /.project
 /cabal.sandbox.config
+tags
 /tags
 /.cabal-sandbox
 /.stack-work
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -35,6 +35,7 @@ data Options = Options
     ,arguments :: [String]
     ,test :: Maybe String
     ,warnings :: Bool
+    ,nostatus :: Bool
     ,height :: Maybe Int
     ,width :: Maybe Int
     ,topmost :: Bool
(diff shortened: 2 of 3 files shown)


src/Ghcid.hs
```diff
@@ -52,6 +53,7 @@ options = cmdArgsMode $ Options
     ,arguments = [] &= args &= typ "MODULE"
     ,test = Nothing &= name "T" &= typ "EXPR" &= help "Command to run after successful loading"
     ,warnings = False &= name "W" &= help "Allow tests to run even with warnings"
+    ,nostatus = False &= name "S" &= help "Suppress status messages"
```
ndmitchell: I suspect you also want `& name "no-status"` here, or to rename the field `no_status`, so that it gets the flag `--no-status` rather than `--nostatus`.

ndmitchell:
Looking at https://travis-ci.org/ndmitchell/ghcid/jobs/129419778 I see:

```
src/Test/Polling.hs:45:20:
    Couldn't match expected type ‘IO ()’
                with actual type ‘([(Ghcid.Style, String)] -> IO ()) -> IO ()’
```

I suspect you need to add that extra `Bool` to the tests as well. Other than that, with the line comment, looks good to me.

.gitignore
```diff
@@ -2,6 +2,7 @@
 /.dist-buildwrapper
 /.project
 /cabal.sandbox.config
+tags
 /tags
```
ndmitchell: Ignoring both `tags` and `/tags` seems redundant. Can you delete the `/tags`.

ndmitchell:
All merged, thanks a lot. Would putting out a release with this change be of benefit to you? If not, I'll leave it for the next changes to go in too and release it then.

pkamenarsky:
Glad I could help! I'm using a local build for the time being, so don't bother putting out a release right now.

ndmitchell:
I actually don't think I've got any changes planned any time soon, so I made a release now with it.

PR #82, opened 2016-08-05
parse correctly messages with spans

JPMoresmau:
When -ferrors-spans is set, the message parsing does not include the end column in the position parsing, which means warnings are not properly identifier and are marked as errors. This fixes it (with test added to prove it!).


ndmitchell:
Awesome work! Thanks a lot.

One remark, all your tests have the format `file:line:col1-col2`. Does GHC ever put `file:line1-line2:col1-col2`? Or does it do something like `file:line1:col1-line2:col2`? Whichever it does, it would be good to have one of the span tests checking that as well.

PR #83, opened 2016-08-05
Multi line spans

JPMoresmau:
Following your remark, I found an error with a mutliple line span, which is now handled.


ndmitchell:
Awesome, thanks a lot.

PR #86, opened 2016-09-03
Merge neovim-ghcid into ghcid

cloudhead:
This adds a top-level `plugins/` directory as the home for editor plugins and merges the neovim-ghcid codebase into `plugins/nvim`.

To get an idea of what this would look like, you can browse it here: https://github.com/cloudhead/ghcid

It would also be possible to include all the integrations at the root level, like in https://github.com/nsf/gocode, but I prefer this approach as it's a little cleaner. This also lets us have a separate readme for plugin-related things, that we can point to from the top-level readme.


ndmitchell:
Thanks! That looks great. From what I can tell, there's no requirement to make a release to include your new work - everything in neovim is sucking it straight out of the repo. Can you confirm that?

cloudhead:
Exactly! It's also *possible* to have the plugin manager install ghcid automatically when the plugin is installed. I'll probably look into that at some point.

PR #103, opened 2017-08-22
Ghc 8.2 formats error messages differently

asivitz:
Previously, messages would be entirely whitespace indented.
Now, one of the indented lines may be prefixed by a line number, which
should no longer be recognized as the start of a new message.

### DIFF
--- a/src/Language/Haskell/Ghcid/Parser.hs
+++ b/src/Language/Haskell/Ghcid/Parser.hs
@@ -40,7 +40,7 @@ parseLoad  = nubOrd . f
             , (pos,rest) <- span (\c -> c == ':' || c == '-' || isSpan c || isDigit c) rest
             -- separate line and column, ignoring span (we want the start point only)
             , [p1,p2] <- map read $ wordsBy (\c -> c == ':' || isSpan c) $ takeWhile (/= '-') pos
-            , (msg,las) <- span (isPrefixOf " ") xs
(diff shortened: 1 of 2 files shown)


src/Language/Haskell/Ghcid/Parser.hs
```diff
@@ -58,6 +58,11 @@ parseLoad  = nubOrd . f
         f [] = []
         isSpan c = c== ',' || c == '(' || c == ')'
 
+-- After the file location, message bodies are indented (perhaps prefixed by a line number)
+isMessageBody :: String -> Bool
+isMessageBody xs = isPrefixOf " " xs || isPipeIndented xs
+  where isPipeIndented = all (\x -> isSpace x || isDigit x) . takeWhile (/= '|')
```
ndmitchell: This doesn't require a pipe character - the line "12" matches for example. Should there be a requirement to find the pipe?

asivitz: Yea that would probably be better in the long term; I'll update it tonight

asivitz:
Not sure if you saw but I updated this PR

ndmitchell:
I didn't! Thanks for pointing it out - now merged. Do you think it's making a release specifically for this change?

asivitz:
I guess I do; I'm no expert on the GHC 8.2 changes but ghcid wouldn't work for me on GHC 8.2.1 without this PR.

PR #104, opened 2017-08-24
Add ignore-loaded option.

chpatrick:
In our project, we need to use `-fobject-code` with GHCi. This means that if the project has already been built, GHCi loads instantly without loading any files, which causes `ghcid` to stop. This option lets you ignore that as long as you specify `--reload` yourself.


ndmitchell:
Makes a lot of sense. We might be able to do better for object code cases in future, but this at least seems a short term workaround.

PR #106, opened 2017-10-02
[emacs] Initial support

WraithM:
Here's super basic emacs support for ghcid. This uses a terminal with `compilation-mode`.

This code also assumes that stack is installed, which maybe isn't a great assumption. I had to do that in the past to get ghcid to behave properly with my stack environment, but perhaps ghcid works well now with stack without having to invoke the stack command directly.


ndmitchell:
Thanks! I also copied and pasted your PR message into the readme, which will at least give some help to people who come across it.

PR #107, opened 2017-10-03
update vscode plugin README

yairchu:
Added steps I needed to successfully perform a local installation over here -

* Installing `vsce` and using it from installed location
* Outdated plugin version number. Replaced with `*` to require less maintainence.

### DIFF
--- a/plugins/vscode/README.md
+++ b/plugins/vscode/README.md
@@ -20,5 +20,6 @@ Requires ghcid and a lot of tedious manual setup (the steps above).
 Run:
 
     npm install
-    vsce package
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks!

ndmitchell:
Are these the right steps? I just did `npm install -g vsce` and that got `vsce` on my PATH, which seems to be recommended here: https://code.visualstudio.com/docs/extensions/publish-extension

ndmitchell:
Without some tweaks it was failing for me, so have gone with `-g` and not an explicit path, which should hopefully work for everyone.

yairchu:
Cool. With `-g` works fine here.

PR #108, opened 2017-10-03
vscode plugin handles GHC 8.2 error messages

yairchu:
Similar to https://github.com/ndmitchell/ghcid/pull/103 -
GHC 8.2's error messages have changed and the vscode plugin needs to be updated or it doesn't properly produce the errors.

An alternative to this approach would be to leave the plugin unchanged but modify `ghcid`'s output so it looks different than GHC's in that the errors are all indented.

### DIFF
--- a/plugins/vscode/package.json
+++ b/plugins/vscode/package.json
@@ -2,7 +2,7 @@
     "name": "haskell-ghcid",
     "displayName": "haskell-ghcid",
     "description": "Integrate ghcid into VS Code",
-    "version": "0.1.0",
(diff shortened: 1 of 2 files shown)


ndmitchell:
Awesome! Thanks for the fix. I'd say that if we want ghcid to do more, it should be producing JSON to ghcid doesn't need to do anything - but I deliberately tried to do the plugin without changing ghcid to start with.

ndmitchell:
I published 0.1.1 with your change

PR #111, opened 2017-11-10
More helpful error message for no files loaded

### LINKED ISSUE #110: Add warning to the "No files loaded, GHCi is not working properly" error message
Suggest that something might be suppressing module output such as `:set -v0` in the arguments or in their `.ghci` config files. Example module output so they know what `ghci` / `stack ghci` looks like would be good too.

bitemyapp:
screenshot from 2017-11-10 15-23-15

Fixes #110

I can go into more detail about the debugging session but this was pretty gnarly to figure out, especially if you haven't been bitten by `.ghci` files before. I had had problems once before when trying to get GHCJS to bootstrap.

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -251,7 +251,14 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
     nextWait <- waitFiles waiter
     (messages, loaded) <- sessionStart session command
     when (null loaded && not ignoreLoaded) $ do
-        putStrLn $ "\nNo files loaded, GHCi is not working properly.\nCommand: " ++ command
+        putStrLn $ "\nNo files loaded, GHCi is not working properly.\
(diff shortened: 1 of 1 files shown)


ndmitchell:
The alternative would be to put `:set -v1` or similar at the start of the ghcid execution - would that work? Specifically at https://github.com/ndmitchell/ghcid/blob/master/src/Language/Haskell/Ghcid.hs#L162, where we disable other ghci customisations.

ndmitchell:
PS. Nice debugging, that doesn't sound easy!

bitemyapp:
>The alternative would be to put :set -v1 or similar at the start of the ghcid execution - would that work? Specifically at https://github.com/ndmitchell/ghcid/blob/master/src/Language/Haskell/Ghcid.hs#L162, where we disable other ghci customisations.

Ah yes! That's a vast improvement on just changing the error message! Thank you!

Issue #43, opened 2015-07-03
break-on-exception causes problems


ndmitchell:
See https://www.reddit.com/r/haskell/comments/3bqy5h/survey_which_haskell_development_tools_are_you/csqxz2s for the full details. They have a .ghci with:

    :set -fbreak-on-exception
    :set -fbreak-on-error

Ghcid raises an error to flush its buffers. It should turn off those flags.

ndmitchell:
Done, and works nicely.

PR #112, opened 2017-11-10
Always set verbosity to 1

### LINKED ISSUE #110: Add warning to the "No files loaded, GHCi is not working properly" error message

bitemyapp:
Fixes #110 instead of #111 as it is better to just prevent the problem.

### DIFF
--- a/src/Language/Haskell/Ghcid.hs
+++ b/src/Language/Haskell/Ghcid.hs
@@ -159,7 +159,7 @@ startGhci cmd directory echo0 = do
                 writeIORef stderr []
                 writeInp "import qualified System.IO as INTERNAL_GHCID"
                 writeInp $ ":set prompt " ++ ghcid_prefix
-                writeInp ":set -fno-break-on-exception -fno-break-on-error" -- see #43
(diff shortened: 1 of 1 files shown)


src/Language/Haskell/Ghcid.hs
```diff
@@ -159,7 +159,7 @@ startGhci cmd directory echo0 = do
                 writeIORef stderr []
                 writeInp "import qualified System.IO as INTERNAL_GHCID"
                 writeInp $ ":set prompt " ++ ghcid_prefix
-                writeInp ":set -fno-break-on-exception -fno-break-on-error" -- see #43
+                writeInp ":set -v1 -fno-break-on-exception -fno-break-on-error" -- see #43
```
ndmitchell: Can you adjust the comment to say see #43 and #110, so we can understand the lineage of -v1?

bitemyapp: Absolutely, good idea.

ndmitchell:
Thanks!

PR #117, opened 2017-12-11
improve vscode extension README

yairchu:
current instructions may cause to install older locally built versions (my bad..) that were left in the folder


ndmitchell:
Sounds like a good idea. Given this is for users though, not scripts, should it just be `rm haskell-ghcid-*.vsix` ?

yairchu:
Yeah sounds reasonable :) Done

ndmitchell:
Thanks!

PR #118, opened 2017-12-11
Remove unnecessary Watch Ghcid output command

yairchu:
Since the vscode package is no longer requires the old watching output file thing,
the extra command now is just confusion prone.

### DIFF
--- a/plugins/vscode/package.json
+++ b/plugins/vscode/package.json
@@ -16,16 +16,11 @@
     },
     "categories": ["Languages", "Linters"],
     "activationEvents": [
-        "onCommand:extension.watchGhcidOutput",
         "onCommand:extension.startGhcid"
     ],
     "main": "./out/src/extension",
     "contributes": {
         "commands": [
-            {
-                "command": "extension.watchGhcidOutput",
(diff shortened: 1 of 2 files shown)


ndmitchell:
I think the Watch output is still useful. Imagine you have a large wrapper script that does lots of things, including setting up `ghcid` and potentially installing it, then you might want to use that script as the starting command for `ghcid`, but still watch the output. Does that make sense?

ndmitchell:
Of course, if it's still confusion prone, renaming it might be wise.

yairchu:
In the specific example described, wouldn't you build `ghcid` once and then you could just use the existing command?

ndmitchell:
In a work environment often the stack resolver gets upgraded regularly (e.g. weekly) and you'd have to rebuild `ghcid` each time. An alternative would be to have a setting so people can set their own `ghcid` command as a VS Code setting, and then people who needed it could do it that way (and it would still be easier than today, and delete the watch command).

yairchu:
Alright now made the `ghcid` command configurable.

ndmitchell:
Looks great!

PR #122, opened 2017-12-21
oops reorder steps in README

yairchu:
My previous README patch accidentally mixed the steps.. This order is right and verified..


ndmitchell:
Good catch

Issue #124, opened 2017-12-27
Add timestamp to "All good" message


yairchu:
Being used to things not working, when I see the "All good" message I'm never 100% sure that all is really good, or perhaps `ghcid` got stuck. While it never got stuck, this doubt I have and I'm sure many others share would be mostly alleviated if it was timestamped -

    All good (224 modules, at 13:06)

ndmitchell:
Sounds like a great idea. PR most welcome!

ndmitchell:
Done by @vaibhavsagar in #125.

PR #125, opened 2017-12-29
Add timestamp to "All good" message

vaibhavsagar:
Here's a quick first stab at the timestamp behaviour requested in #124. My setup is a bit unusual (NixOS) and I couldn't get the tests running so I don't know if/how the tests need to be updated if at all.

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -186,7 +186,8 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
         outputFill load msg = do
             (width, height) <- termSize
             let n = height - length msg
-            load <- return $ take (if isJust load then n else 0) $ prettyOutput (maybe 0 fst load)
+            currTime <- getShortTime
+            load <- return $ take (if isJust load then n else 0) $ prettyOutput currTime (maybe 0 fst load)
                 [ m{loadMessage = concatMap (chunksOfWord width (width `div` 5)) $ loadMessage m}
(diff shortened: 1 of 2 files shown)


src/Ghcid.hs
```diff
@@ -216,15 +218,15 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
 
             let updateTitle extra = unless no_title $ setTitle $
                     let f n msg = if n == 0 then "" else show n ++ " " ++ msg ++ ['s' | n > 1]
-                    in (if countErrors == 0 && countWarnings == 0 then allGoodMessage else f countErrors "error" ++
+                    in (if countErrors == 0 && countWarnings == 0 then (allGoodMessage ++ " at " ++ currTime) else f countErrors "error" ++
```
hanjoosten: remove these brackets to get hlint happy 👍

vaibhavsagar: Sure!

ndmitchell:
Thanks for the patch! I'll take a look in the next day or so.

yairchu:
Nice! I now use this :)

ndmitchell:
Thanks! I tweaked the code to use the `time` packages `formatTime` function, and included the seconds as well, since that seemed sensible to you know if it's changing or has been frozen for the last minute. Code in 8bc299a.

vaibhavsagar:
Thanks so much!

PR #133, opened 2018-02-05
Add a lower bound to directory

gwils:
directory 1.1.0.2 is a boot library on GHC 7.4, and that version has an incompatible type on `getModificationTime`. This lower bound precludes those bad build plans.

Please also make hackage metadata revisions to add this bound to older releases as necessary: https://matrix.hackage.haskell.org/package/ghcid#GHC-7.4/ghcid-0.6.9


ndmitchell:
Thanks! Will go metadata mangling later today.

ndmitchell:
Metadata now all seems good.

gwils:
Looks beautiful to me! Thanks a lot.

PR #139, opened 2018-02-22
Add option to remove bold escape codes

rjmk:
I've been playing around with using ghcid with plan9's acme editor. It displays some ugly cruft when the build displays bold output.

This adds an option to turn off trying to display bold output.

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -51,6 +51,7 @@ data Options = Options
     ,outputfile :: [FilePath]
     ,ignoreLoaded :: Bool
     ,poll :: Maybe Seconds
+    ,removeBold :: Bool
     }
     deriving (Data,Typeable,Show)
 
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks! I'll review this, probably over the weekend.

ndmitchell:
Thanks, seems like a valuable thing to have, but rather than restrict it to bold, I suggest we follow the same approach as HLint at https://github.com/ndmitchell/hlint/blob/master/src/CmdLine.hs#L76. The essentially names the flag `color`, and tries to default to something sensible if it can.

src/Ghcid.hs
```diff
@@ -263,8 +265,9 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
 
 
 -- | Given an available height, and a set of messages to display, show them as best you can.
-prettyOutput :: String -> Int -> [Load] -> [(Style,String)]
-prettyOutput currTime loaded [] = [(Plain,allGoodMessage ++ " (" ++ show loaded ++ " module" ++ ['s' | loaded /= 1] ++ ", at " ++ currTime ++ ")")]
-prettyOutput _ loaded xs = concat $ msg1:msgs
+prettyOutput :: String -> Int -> Bool -> [Load] -> [(Style,String)]
+prettyOutput currTime loaded _ [] = [(Plain,allGoodMessage ++ " (" ++ show loaded ++ " module" ++ ['s' | loaded /= 1] ++ ", at " ++ currTime ++ ")")]
+prettyOutput _ loaded removeBold xs = concat $ msg1:msgs
```
ndmitchell: Instead of wiring `removeBold` all the way inside `prettyOutput`, I suggest you have `termOutput` take a boolean saying color or not, and remove all styles if it is False.

ndmitchell:
I've pushed a patch that gives `--color=never` to disable the Bold. In particular I'll be allowing more color in the near future, so will make sure `--color=never` keeps suppressing it. Furthermore, if your terminal correctly advertises it doesn't support ANSI then hopefully just doing nothing will still work.

rjmk:
Thanks for fixing this, Neil! Sorry I didn't get a chance to incorporate your suggestions.

I'll be on a setup where I can give acme a go again soon, so I'll be pleased to have this

Issue #134, opened 2018-02-05
Making ghcid more usable by external tools


merijn:
So, ghcid seems to be the only interactive tool that actually works reasonably with `new-build` I'm trying to get it working with vim and neomake (I don't have neovim available, so sadly the existing nvim plugin doesn't quite work). I'd like to propose some minor tweaks that make ghcid slightly more friendly to tooling. I'm already implementing this stuff, so this issue is more to discuss behaviour preferences to get the finished work merged back into ghcid (because I don't wanna keep porting it forward indefinitely).

Desirable changes when reading ghcid output in tools:
* Disable ANSI colouring
* Don't linewrap
* Sentinel output to indicate the end of error/warning list

My personal changes to the design would be as follows (but I'm open to ensuring the default behaviour doesn't change and putting all the above changes behind flags, if you prefer to preserve the current behaviour at all cost):

* Auto-detect when ghcid is talking to an ANSI terminal.
* Disable ANSI colouring when not talking to a terminal (unless a new flag forces colouring on).
* Disable the linewrapping/error truncation when not talking to a terminal (unless an explicit width/height are passed in on the command line).
* Output a sentinel at the end of output when height is not limited. (This is the change I'm most torn about, it might be nicer to just never output a sentinel unless explicitly instructed by a flag...)

While this changes some of the default behaviour of ghcid, it shouldn't affect most usecases, as I think in most cases people use it attached to a terminal anyway. But, as I said, I'm open to preserving the current behaviour as-is and hiding all my changes between flags if that's preferred.

merijn:
After playing around a bit with it, I think a lot of the terminal code is rather deeply intertwined with the other ghcid code so that making these changes becomes rather invasive.

A hacky, but much simpler approach might be to not change any of default behaviours at all and just add the ability to inhibit doing output at all. Because then the desired effect can be achieved by simply running `ghcid` with `-o /dev/stdout`, which already logs all messages without any wrapping  or ANSI colour noise.

merijn:
That would still necessitate adding a flag to output some sentinel line to distinguish when error reports end, but that should be fairly easy. Which would reduce any changes to one flag to write such a sentinel line and another flag to inhibit the standard output, which should also be fairly trivial.

ndmitchell:
Happy to make changes. I guess the first question is how you interact with it - I expected that for downstream tooling the `-o` output mode would be easiest - that's what the VS Code plugin uses. With respect to that:

[...]

PR #142, opened 2018-03-08
Update API to distinguish between shell invocations and executables.

merijn:
See #134

### DIFF
--- a/ghcid.cabal
+++ b/ghcid.cabal
@@ -1,7 +1,7 @@
 cabal-version:      >= 1.18
 build-type:         Simple
 name:               ghcid
-version:            0.6.10
+version:            0.6.11
 license:            BSD3
 license-file:       LICENSE
 category:           Development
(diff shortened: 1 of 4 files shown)


ghcid.cabal
```diff
@@ -1,7 +1,7 @@
 cabal-version:      >= 1.18
 build-type:         Simple
 name:               ghcid
-version:            0.6.10
+version:            0.6.11
```
ndmitchell: As a note for next time (but not really a big deal either way) I only update the version number when I make a release, as per the last bullet at https://github.com/ndmitchell/neil#contributions

ndmitchell:
Thanks! Sorry for taking so long to review...

In the end I decided that someone is going to want environment variables at some point, and all the other process tweaks, so I opted to just expose `startGhciRaw` (renamed to `startGhciProcess`) and the existing `startGhci`, but nothing further.

PR #146, opened 2018-04-07
Add --max-messages/-n flag to limit number of printed messages

waern:
Hi Neil,

thanks for an excellent tool. Here is a patch that adds a `--max-messages` (`-n`) flag that limits the number of messages printed. I use it to print only one single message at a time.


waern:
Doh, I missed the fact that you have a test suite. I'll take a look at that.

waern:
The appveyor build failure seems unrelated to my changes. The test suite passes locally.

ndmitchell:
Thanks! Makes a lot of sense, all looks very good. I'll investigate the test suite...

PR #156, opened 2018-04-15
Support a filetype 'ghcid' in the neovim plugin

aiya000:
(I'm not English native, please tell the right sentence if the written sentence is unnatural :smiley:)

This allows to set the ftplugin by the user side that is like below .vim/after/ftplugin/ghcid.vim :dog2:

```vim
nnoremap <buffer><silent> <localleader>R :<C-u>terminal stack test<CR>
```

The above runs `stack test` :muscle:


ndmitchell:
Thanks for the PR. Ping @cloudhead to weigh in.

PR #166, opened 2018-05-25
[nvim] fix ghcid version requirement in README

mheinzel:
The plugin passes `--color=never` to ghcid, which is only available since version `0.7`


ndmitchell:
Looks reasonable to me - @cloudhead  - any thoughts?

I have a plan for making ghcid allow optional flags so future upgrades don't require a break even if you want to pass new flags.

cloudhead:
Yep, thanks!

PR #179, opened 2018-07-13
(nvim) Add g:ghcid_verbosity option

mitchellwrosen:
This patch adds a `g:ghcid_verbosity` flag that controls how much the plugin echoes.

- 0: echo no "ok" nor "warning" messages (but do indicate if ghcid was killed, or died, etc)
- 1: echo "warning" messages but not "ok"
- 2: echo everything (default)

cc @cloudhead

### DIFF
--- a/plugins/nvim/ftplugin/haskell.vim
+++ b/plugins/nvim/ftplugin/haskell.vim
@@ -29,6 +29,10 @@ if !exists("g:ghcid_background")
   let g:ghcid_background = 0
 endif
 
+if !exists("g:ghcid_verbosity")
+  let g:ghcid_verbosity = 2
+endif
+
 let s:ghcid_command_args = {}
 let s:ghcid_base_sign_id = 100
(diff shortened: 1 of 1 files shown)


mitchellwrosen:
My use case is as follows:

```
function! <SID>EchoQuickFixEntry()
  let l:entries = getqflist()
  let l:bufnr = bufnr('%')
  let l:lnum = line('.')
  for e in entries
    if e.bufnr == l:bufnr && e.lnum == l:lnum
      echo e.text
      return
    endif
  endfor
endfunction
au CursorMoved * cal <SID>EchoQuickFixEntry()
```

This autocommand's output competes with ghcid's output, and often I end up seeing `Ghcid: 1 warning(s)` even when hovering over the line with a warning.

cloudhead:
Looks good.

cloudhead:
Can you rewrite your commit so it's formatted like `[nvim] ...`? Happy to merge then.

mitchellwrosen:
Done!

PR #186, opened 2018-07-27
Add -ignore-dot-ghci to tests, otherwise some fail with erroring .ghci

### LINKED ISSUE #185: Some tests fail when there is erroring ~/.ghci
I have `~/.ghci` which loads a ppr-library, and I constantly forget to install it when switching machines. I noticed that `ghcid`'s tests don't like that. Probably, passing `-ignore-dot-ghci` should cure that not breaking anything (PR is coming). It might be the case, that `ghcid` itself should use this flag by default.

ulysses4ever:
Fix #185.

### DIFF
--- a/src/Test/API.hs
+++ b/src/Test/API.hs
@@ -13,7 +13,7 @@ import Language.Haskell.Ghcid.Util
 apiTests :: TestTree
 apiTests = testGroup "API test"
     [testCase "No files" $ withTempDir $ \dir -> do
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks! In future feel free to just create a PR notifying of the bug and the fix in one, if you prefer (I'm easy either way)

ulysses4ever:
Thanks! One PR is easier, indeed!

PR #188, opened 2018-07-27
Replace map Esc / fromEsc with GHC's coerce

### LINKED ISSUE #187: refactor `map Esc` / `map fromEsc` to `coerce`
There are a couple of places in `Parser.hs` which do `map` on the `Esc` newtype (de)constructor. This is the very use case for GHC's coercion mechanism.

A note on compatibility: the docs say `Data.Coerce` appeared in `base 4.7` which is the current lower bound in `ghcid`'s cabal file.

ulysses4ever:
Fix #187.

### DIFF
--- a/src/Language/Haskell/Ghcid/Parser.hs
+++ b/src/Language/Haskell/Ghcid/Parser.hs
@@ -12,6 +12,7 @@ import Data.Maybe
 import Text.Read
 import Data.Tuple.Extra
 import Control.Applicative
+import Data.Coerce
 import Prelude
 
 import Language.Haskell.Ghcid.Types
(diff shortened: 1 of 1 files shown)


ndmitchell:
I find the `map fromEsc` much clearer to read, and given that it's processing things with `[Char]`, I can't imagine it's a noticable performance difference anyway. My inclination is to retain the clarity and avoid coerce.

ulysses4ever:
Sure, it is up to you. For good or for bad, I did not measure performance. I agree that `map` is clearer. I just can't stand this kind of things that were known to be annoying and were fixed, but still pervade code… Maybe, we need more time to mentally embrace `coerce`.

ulysses4ever:
Thanks a lot for sharing your thoughts on this. They are illuminating to me.

PR #190, opened 2018-07-30
Load benchmarks as well as tests

Bodigrim:
This is also more consistent with [ghcid.el](https://github.com/ndmitchell/ghcid/blob/master/plugins/emacs/ghcid.el#L9).


ndmitchell:
Thanks! Good catch.

Issue #196, opened 2018-08-22
--verbose should print out the version number


ndmitchell:
Then I have all the info I need in one place

Issue #172, opened 2018-06-26
--verbose should print out the version


ndmitchell:
So people who attach a complete verbose listing are providing all the information.

Also System.Info.os and arch.

PR #202, opened 2018-09-08
add os, version, and architecture to verbose output

mchowning:
Adding System.Info.os, System.Info.arch, and the ghcid version to the verbose output. Addressing #172 and #196.

```
%OS: darwin
%ARCH: x86_64
%VERSION: 0.7
```


ndmitchell:
Thanks!

PR #225, opened 2019-02-26
Watching directories

brprice:
Two minor tweaks to `--reload`:
1) `--reload .` does not work (as "." is erroneously considered hidden). (We also now consider ".." to be not hidden.)
2) Explicitly wait on directories, to pick up new files (this has only been tested under Linux)

### DIFF
--- a/src/Wait.hs
+++ b/src/Wait.hs
@@ -32,6 +32,16 @@ withWaiterNotify f = withManagerConf defaultConfig{confDebounce=NoDebounce} $ \m
     var <- newVar Map.empty
     f $ WaiterNotify manager mvar var
 
+-- `listContentsInside test dir` will list files and directories inside `dir`,
+-- recursing into those subdirectories which pass `test`.
+-- Note that `dir` and files it directly contains are always listed, regardless of `test`.
+-- Subdirectories will have a trailing path separator, and are only listed if we recurse into them.
+listContentsInside :: (FilePath -> IO Bool) -> FilePath -> IO [FilePath]
+listContentsInside test dir = do
+    (dirs,files) <- partitionM doesDirectoryExist =<< listContents dir
+    recurse <- filterM test dirs
+    rest <- concatMapM (listContentsInside test) recurse
+    return $ addTrailingPathSeparator dir : files ++ rest
 
 -- | Given the pattern:
 --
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks for the patch - interesting! So the two changes are:

* Add the directories explicitly to the list of things to watch. Seems very reasonable. Do you have to put the trailing slash on the end? If not, I'd rather we didn't - since it's implying a semantic difference that isn't there.
* Allowing `.` as the argument. An interesting approach, but I wonder if there's a simpler way around. At the moment with `listContentsInside` we check the predicate, then descend if it passes. However, even if the user passes `.git` as their initial folder, they probably do want to go inside it there. Therefore, how about keeping the original predicate, but changing the function to check after, so whatever the user passes initially will escape scrutiny?

Both seem good ideas.

brprice:
- The trailing slash because when creating the directories to `notify` on, we use `map takeDirectory files`, but some of the files are actually directories. Without a trailing slash this would actually wait on the parent directory. This could be cooked some other way if you prefer, though I guess returning directories and files separately from `listContentsInside` still implies that semantic difference, and I'd prefer not to make an IO call to check if it is a file or directory. What do you suggest?

- Good point, that does sound better.

ndmitchell:
OK, you point about `/` makes a lot of sense. But now I'm wondering if waiting for a directory really works? The way the wait code works is by waiting the watcher, and then checking all the files it cares about and seeing if any have modification time changes. I suspect if it is working for you, that's because the modification time thing fails entirely, and then you're going down the failure path of give up and reload anyway. I don't disagree it is a good idea though!

brprice:
I'm not following you here, sorry. Looking at the loop in `recheck`, I don't see where the give-up-and-reload-anyway path is.

In any case, I'm not seeing any difference between files and directories: (snippets from verbose output, file first, then directory)

```
%NOTIFY: Modified "/home/ben/dev/ghcid/test-reload-dir/file" 2019-02-28 17:54:55.859643896 UTC False
%WAITING: Notify signaled 
%RELOADING: test-reload-dir/file

Reloading...                      
  test-reload-dir/file
```

```
%NOTIFY: Unknown "/home/ben/dev/ghcid/test-reload-dir/" 2019-02-28 17:56:24.521966815 UTC "Attributes {isDirectory = True, maybeFilePath = Nothing}" 
%NOTIFY: Modified "/home/ben/dev/ghcid/test-reload-dir" 2019-02-28 17:56:24.521954763 UTC True
%WAITING: Notify signaled
%RELOADING: test-reload-dir/

Reloading...                      
  test-reload-dir/
```
(As an aside, it does seem a bit silly that the notifier tells us what file it was notifing about, only for us to throw the information away and poll all files for modification times!)
(Another aside, the logic for polling seems broken: we sleep the requested time on the first iteration (in `waitFiles`, but unless there was a change, we then loop in `recheck` and this sleeps for only `0.1` seconds)

ndmitchell:
To address your remarks:

* There is no give up and reload branch, my bad.
* The loop through modification times is for safety - we might get spurious notifies. The simplest thing was just to check everything, and it should be on a relatively small number of files, so not a big deal.
* What I don't see is how new files will escape recheck - won't they not change because getModTime doesn't see them so will just call `recheck` again?

brprice:
Ah, I think I see what you mean. Indeed new files will be unseen by recheck, but the _directory_'s modTime changes when a new file is created inside it, and this is what triggers escaping recheck (indeed, this is the point of adding directories to the watchlist).

ndmitchell:
I hadn't thought of the modtime on a directory - OK - that makes sense then. Can we get a comment to that effect somewhere, otherwise I'm going to forget again.

Happy to merge once we switch to checking the predicate after listing rather than before.

brprice:
Done and done. I think this is ready to merge now.

ndmitchell:
I agree! Thanks a lot.

ndmitchell:
Would you like a release with this change incorporated? I'm happy either way (otherwise it will go in when I next have a larger batch of changes).

ndmitchell:
Actually, I think there's enough in to justify a release, so doing one now.

PR #236, opened 2019-04-09
Add hlint support

### LINKED ISSUE #158: Run HLint on each file
For each file, it should run HLint on the file if it can and integrate the output.

jonascarpay:
Closes #158.
This is a little WIP proof of concept of hlint integration. It adds a `-l` flag that will run hlint if there are no errors, which covers my use cases. 
- [ ] On a 20K line project, this takes about 3s on my machine, is that reason to turn hlint into a persistent process?
- [ ] Do we want more command-line options than `-l`? I should add `.hlint.yaml` support at least.
- [ ] Are you happy with this approach @ndmitchell, or am I overlooking a more idiomatic way to do this?

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -22,6 +22,7 @@ import System.Directory.Extra
 import System.Time.Extra
 import System.Exit
 import System.FilePath
+import System.Process (readProcessWithExitCode)
 import System.Info
 import System.IO.Extra
 
@@ -30,7 +31,6 @@ import Language.Haskell.Ghcid.Escape
 import Language.Haskell.Ghcid.Terminal
 import Language.Haskell.Ghcid.Util
 import Language.Haskell.Ghcid.Types
-import Language.Haskell.HLint3 (hlint)
 import Wait
 
 import Data.Functor
@@ -44,7 +44,7 @@ data Options = Options
     ,test :: [String]
     ,run :: [String]
     ,warnings :: Bool
-    ,lint :: Bool
+    ,lint :: Maybe String
     ,no_status :: Bool
     ,height :: Maybe Int
     ,width :: Maybe Int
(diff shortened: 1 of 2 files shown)


ndmitchell:
One of the nice things about `ghcid` is it has only light bindings to other tools like ghci. I wonder whether simply shelling out to `hlint` is the more `ghcid` way to go? That would also give things like `.hlint.yaml` support for free.

I think the approach you've taken, nice and simple, is a good one. A much better route than trying something more complex.

jonascarpay:
Maybe this is better? I changed it so that you can do `-l=LINTCMD`, which defaults to hlint. This should allow you to use other linters as well.

src/Ghcid.hs
```diff
@@ -75,6 +77,7 @@ options = cmdArgsMode $ Options
     ,test = [] &= name "T" &= typ "EXPR" &= help "Command to run after successful loading"
     ,run = [] &= name "r" &= typ "EXPR" &= opt "main" &= help "Command to run after successful loading (defaults to main)"
     ,warnings = False &= name "W" &= help "Allow tests to run even with warnings"
+    ,lint = Nothing &= typ "LINTCMD" &= name "lint" &= opt "hlint" &= help "Run linter if there are no errors. Defaults to hlint."
```
ndmitchell:
```suggestion
    ,lint = Nothing &= typ "COMMAND" &= name "lint" &= opt "hlint" &= help "Linter to run if there are no errors. Defaults to hlint."
```

src/Ghcid.hs
```diff
@@ -326,6 +330,10 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
                      else do
                         updateTitle "(test done)"
                         whenNormal $ outStrLn "\n...done"
+            whenJust lint $ \lintcmd ->
+                unless hasErrors $ do
+                    (exitcode, stdout, _) <- readProcessWithExitCode lintcmd loaded ""
```
ndmitchell:
```suggestion
                    (exitcode, stdout, stderr) <- readProcessWithExitCode lintcmd loaded ""
```

src/Ghcid.hs
```diff
@@ -326,6 +330,10 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
...
+                    unless (exitcode == ExitSuccess) $ outStrLn stdout
```
ndmitchell:
```suggestion
                    unless (exitcode == ExitSuccess) $ outStrLn $ stdout ++ stderr
```

src/Ghcid.hs
```diff
@@ -22,6 +22,7 @@ import System.Directory.Extra
 import System.Time.Extra
 import System.Exit
 import System.FilePath
+import System.Process (readProcessWithExitCode)
```
ndmitchell:
```suggestion
import System.Process
```

ndmitchell:
I made a few little tweaks, hoping to apply them myself and then merge, but it seems I've done them in such a way that only you can apply. If you think they make sense can you do so and then I'll merge.

ndmitchell:
The direction of `--lint` is very smart, makes it nice and reusable and fits very well with the ghcid approach.

jonascarpay:
Thanks for the quick response! I've applied the suggestions, let me know if you want me to add docs/squash/etc.

ndmitchell:
Thanks a lot for this feature, all merged. I'll probably make a release tomorrow.

I3ck:
Nice :)

ndmitchell:
Release made as 0.7.3

PR #237, opened 2019-04-16
Ability to scroll through the error messages

basile-henry:
**Motivation**:
At work we often have to deal with big error messages. We love `ghcid` and use it all the time, but when we get to the big error messages, we end up having to kill `ghcid` and run `cabal new-repl` or any other appropriate command in order to see the full error message and not just the cropped version. This is a much slower iteration loop.

I have searched through the `ghcid` options and the GitHub issues and couldn't find anything that could help with this.

**Solution**:
I understand `ghcid`'s tag line of being a *low feature* IDE. As such I don't add actual scrolling to `ghcid` but rather add a small set of options that make it possible to use the terminal itself or `tmux` to do the scrolling.

A way to output all the errors and warning while still keeping the most relevant ones visible in a terminal is simply to reverse their order.
GHC already offers a flag to do just that: `-freverse-errors` (it works for warnings too).

This PR focuses on adding the following features as 3 separate flags:

- `no-height-limit`: This is the core of the PR, in order to see all the errors we need to disable the height limiting code.
- `reversed`: Uses GHC's `-freverse-errors`. When using the `-c` flag, it's up to the user to call GHC's `-freverse-errors` (either in a `.cabal` file or explicitly in the command). `ghcid` already had some custom logic for ordering errors and warning, this flags changes that custom logic to work properly in reversed order.
- `clear`: When scrolling up we want to see clear divisions between reloads, clearing the screen in between reloads is a nice way to achieve that.

These 3 flags mostly make sense together, so if you would prefer these options were all behind a single flag, I can do that! :smile:

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -216,20 +216,18 @@ mainWithTerminal termSize termOutput =
                   then termSize { termHeight = Nothing }
                   else termSize
 
-                supportsANSI <- hSupportsANSI stdout
-
                 restyle <- do
-                    let useStyle = case color opts of
-                          Always -> True
-                          Never -> False
-                          Auto -> supportsANSI
+                    useStyle <- case color opts of
+                        Always -> return True
+                        Never -> return False
+                        Auto -> hSupportsANSI stdout
                     when useStyle $ do
                         h <- lookupEnv "HSPEC_OPTIONS"
                         when (isNothing h) $ setEnv "HSPEC_OPTIONS" "--color" -- see #87
                     return $ if useStyle then id else map unescape
 
                 clear <- return $
-                  if clear opts && supportsANSI
+                  if clear opts
                   then (clearScreen *>)
                   else id
 
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -46,6 +46,7 @@ data Options = Options
     ,warnings :: Bool
     ,lint :: Maybe String
     ,no_status :: Bool
+    ,clear :: Bool
     ,reversed :: Bool
     ,no_height_limit :: Bool
     ,height :: Maybe Int
@@ -81,6 +82,7 @@ options = cmdArgsMode $ Options
     ,warnings = False &= name "W" &= help "Allow tests to run even with warnings"
     ,lint = Nothing &= typ "COMMAND" &= name "lint" &= opt "hlint" &= help "Linter to run if there are no errors. Defaults to hlint."
     ,no_status = False &= name "S" &= help "Suppress status messages"
+    ,clear = False &= name "clear" &= help "Clear screen when reloading"
     ,reversed = False &=name "reversed" &= help "Reverse output order"
     ,no_height_limit = False &= name "no-height-limit" &= help "Disable height limit"
     ,height = Nothing &= help "Number of lines to use (defaults to console height)"
@@ -214,18 +216,25 @@ mainWithTerminal termSize termOutput =
                   then termSize { termHeight = Nothing }
                   else termSize
 
+                supportsANSI <- hSupportsANSI stdout
+
                 restyle <- do
-                    useStyle <- case color opts of
-                        Always -> return True
-                        Never -> return False
-                        Auto -> hSupportsANSI stdout
+                    let useStyle = case color opts of
+                          Always -> True
+                          Never -> False
+                          Auto -> supportsANSI
                     when useStyle $ do
                         h <- lookupEnv "HSPEC_OPTIONS"
                         when (isNothing h) $ setEnv "HSPEC_OPTIONS" "--color" -- see #87
                     return $ if useStyle then id else map unescape
 
+                clear <- return $
+                  if clear opts && supportsANSI
+                  then (clearScreen *>)
+                  else id
+
                 maybe withWaiterNotify withWaiterPoll (poll opts) $ \waiter ->
-                    runGhcid session waiter (return termSize) (termOutput . restyle) opts
+                    runGhcid session waiter (return termSize) (clear . termOutput . restyle) opts
 
 
 
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -173,7 +173,7 @@ withGhcidArgs act = do
 
 data TermSize = TermSize
     {termWidth :: Int
-    ,termHeight :: Maybe Int
+    ,termHeight :: Maybe Int -- ^ Nothing means the height is unlimited
     ,termWrap :: WordWrap
     }
 
@@ -210,9 +210,9 @@ mainWithTerminal termSize termOutput =
                             (if isJust w then WrapHard else termWrap term)
 
                 termSize <- return $
-                  if no_height_limit opts
-                  then termSize { termHeight = Nothing }
-                  else termSize
+                    if no_height_limit opts
+                    then termSize { termHeight = Nothing }
+                    else termSize
 
                 restyle <- do
                     useStyle <- case color opts of
@@ -225,9 +225,9 @@ mainWithTerminal termSize termOutput =
                     return $ if useStyle then id else map unescape
 
                 clear <- return $
-                  if clear opts
-                  then (clearScreen *>)
-                  else id
+                    if clear opts
+                    then (clearScreen *>)
+                    else id
 
                 maybe withWaiterNotify withWaiterPoll (poll opts) $ \waiter ->
                     runGhcid session waiter (return termSize) (clear . termOutput . restyle) opts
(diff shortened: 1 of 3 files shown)


src/Ghcid.hs
```diff
@@ -33,7 +34,6 @@ import Language.Haskell.Ghcid.Util
 import Language.Haskell.Ghcid.Types
 import Wait
 
-import Data.Functor
```
basile-henry: Old GHCs might not like this

ndmitchell: I have plenty of CI tests for that kind of thing :)

src/Ghcid.hs
```diff
@@ -192,29 +200,41 @@ mainWithTerminal termSize termOutput =
...
+                    let useStyle = case color opts of
+                          Always -> True
+                          Never -> False
+                          Auto -> supportsANSI
                     when useStyle $ do
                         h <- lookupEnv "HSPEC_OPTIONS"
                         when (isNothing h) $ setEnv "HSPEC_OPTIONS" "--color" -- see #87
                     return $ if useStyle then id else map unescape
 
+                clear <- return $
```
ndmitchell: They wrote `--clear`. If their terminal thinks it doesn't support ANSI, I imagine they still want it to clear anyway. I'd skip the supportsANSI check here.

src/Ghcid.hs
```diff
@@ -307,7 +331,10 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
                 -- sort error messages by modtime, so newer edits cause the errors to float to the top - see #153
                 errTimes <- sequence [(x,) <$> getModTime x | x <- nubOrd $ map loadFile msgError]
                 let f x = lookup (loadFile x) errTimes
-                return $ sortOn (Down . f) msgError ++ msgWarn
+                return $
+                  if reversed
+                  then msgWarn ++ sortOn f msgError
+                  else sortOn (Down . f) msgError ++ msgWarn
```
ndmitchell: This doesn't reverse the warnings. I'd rather `(if reversed then reverse else id)` - that way they'll never go out of sync.

src/Ghcid.hs
```diff
@@ -192,29 +200,41 @@ mainWithTerminal termSize termOutput =
...
 
+                termSize <- return $
+                  if no_height_limit opts
+                  then termSize { termHeight = Nothing }
+                  else termSize
+
+                supportsANSI <- hSupportsANSI stdout
+
                 restyle <- do
-                    useStyle <- case color opts of
```
ndmitchell: I'd rather not even query hSupportsANSI unless we have to, since it's the kind of thing I can imagine goes wrong in odd circumstances. Suggest using `once` to capture its value, and then use it in the two places (if needed). But see the remark below.

src/Ghcid.hs
```diff
@@ -167,7 +175,7 @@ withGhcidArgs act = do
 
 data TermSize = TermSize
     {termWidth :: Int
-    ,termHeight :: Int
+    ,termHeight :: Maybe Int
```
ndmitchell: Does this really have to be a Nothing, with all the complexity in logic that implies? Could we just use 100000 if someone says unlimited height? I suspect its close enough that no one is going to have a build that terminates in any reasonable time.

src/Ghcid.hs
```diff
@@ -135,7 +141,9 @@ autoOptions o@Options{..}
         stack <- firstJustM findStack [".",".."] -- stack file might be parent, see #62
 
         let cabal = map (curdir </>) $ filter ((==) ".cabal" . takeExtension) files
-        let opts = ["-fno-code" | null test && null run] ++ ghciFlagsRequired ++ ghciFlagsUseful
+        let opts = ["-fno-code" | null test && null run]
+                ++ ["-freverse-errors" | reversed ]
```
ndmitchell: Is this a good idea? We reverse the errors later on, so if we get GHC to also reverse them aren't we reversing and unreversing the errors - thus putting them back approximately as they were originally?

src/Ghcid.hs
ndmitchell: My CI will know :)

src/Ghcid.hs
basile-henry: My understanding is that this `sortOn` only sorts at the module level. All the messages in a module will have the same timestamp. We still rely on GHC to reverse both errors and warnings within a given module. But maybe I understood this bit wrong?

src/Ghcid.hs
basile-henry: There is some padding logic when the height is defined. I'm not sure we want to print that many empty lines on each reload.
Maybe instead of setting the `termHeight` to be `Maybe`, I could thread the `no_height_limit` option to the relevant places and override it there?

ndmitchell: Ah, I forgot, it's exact height, not maximum. Ignore my remarks - your code is right.

src/Ghcid.hs
basile-henry: As I mentioned we are not reversing them later, just the module order. But maybe I could reverse them later and not use GHC's `-freverse-error` at all. That would probably be a nicer user experience for the `--reversed` with the `-c` flag case (since we wouldn't have to specify `--ghc-options=-freverse-errors` anymore).

My reasoning was that GHC would do a better job at reversing the error messages. But if `ghcid` already has logic to split messages then it might be simpler to roll out our own reversing logic.

src/Ghcid.hs
ndmitchell: True - stable sort. I guess I'd still just, for simplicity have reversed just call reverse, and then it's simple and guaranteed correct. Or are you hoping reverse doesn't reverse in some situations?

basile-henry:
I think leaving the flags separate with no implication like `--reversed` implies `--no-height-limit` is fine. My main worry is that someone will find `--reversed`, try it and find it unusable without thinking about using it with another flag. I can had a hint about using it with the other flag in the help description

I might rename `--reversed` to `reverse-errors` so that it's the same flag as GHC.

I'm not sure about the `--no-height-limit` naming. One of my coworker suggested using `--height 0` for that behaviour. This would have the advantage of getting rid of the `Maybe`s for `termHeight`. But I'm not sure this would be very discoverable feature.

ndmitchell:
`reverse-errors` seems fine. I conisdered `height 0`, but it's also not great as `termHeight` - it really should be Nothing there. So now the absence of a height is something, but 0 = Nothing - that gets confusing quickly.

ndmitchell:
Let me know when you think the code is ready and I'll take another look.

basile-henry:
Here are  some suggestions for a better name for `--no-height-limit`:

- `--no-height`
- `--flexible-height`
- `--dynamic-height`

But otherwise, if you're happy with the current names, I think that's all!

Should I add some examples to the README or other places? The `--help` menu might be enough.

Thank you for the speedy review! :smile:

ndmitchell:
An entry in the FAQ would be good. Along the lines of "What if the error message is too big for my console?" and then with the answer saying trying what you've done in the PR.

src/Ghcid.hs
ndmitchell: Can we get a comment that Nothing means the height is unlimited.

src/Ghcid.hs
```diff
@@ -192,17 +198,22 @@ mainWithTerminal termSize termOutput =
...
                         -- if we write to the final column of the window then it wraps automatically
                         -- so putStrLn width 'x' uses up two lines
                         return $ TermSize
                             (fromMaybe (pred $ termWidth term) w)
-                            (fromMaybe (termHeight term) h)
+                            (h <|> termHeight term)
                             (if isJust w then WrapHard else termWrap term)
 
+                termSize <- return $
+                  if no_height_limit opts
```
ndmitchell: Can this have 4 space indents, like the other code in this block.

src/Ghcid.hs
```diff
@@ -247,9 +263,13 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
                 Just (loadedCount, msgs) -> prettyOutput currTime loadedCount $ filter isMessage msgs
             TermSize{..} <- termSize
             let wrap = concatMap (wordWrapE termWidth (termWidth `div` 5) . Esc)
-            (termHeight, msg) <- return $ takeRemainder termHeight $ wrap msg
-            (termHeight, load) <- return $ takeRemainder termHeight $ wrap load
-            let pad = replicate termHeight ""
+            (msg, load, pad) <-
+              case termHeight of
```
ndmitchell: 4 space indent please

src/Ghcid.hs
basile-henry: Actually, I was just thinking. Is the `termWidth` option only there to control the line wrapping logic, which in turn affects the height of the output?

If that's the case, when we have no height limit we probably don't need a width limit either. This would let the terminal/`tmux` handle the wrapping logic, which has the benefit of not needing a reload of `ghcid` to update the line wrapping.

This is probably out of the score of this PR, but updated line wrapping without reload would be nice 😁

basile-henry:
I'm away from my computer. I'll add the FAQ entry and fix comments/whitespace first thing tomorrow!

ndmitchell:
Awesome! All merged. Anything you want in a close follow up, or should I make a release with this?

basile-henry:
Great thanks! :smile: 

I don't know how often you release a new version of `ghcid`. We might have another pull request coming in a few days about filtering error messages. But it's a fairly different feature and we're not too sure about how it should work just yet, or if we are definitely going to try to upstream that.

About the features in this pull request, I just had another thought. I was focusing on the features working well together, but maybe I can make `--reverse-errors` work slightly better when used on its own. It's not the way I intend on using this flag but just in case others want to use it by itself, I could change the height cropping logic to crop from the other side (top instead of bottom) and also pad on that other side. This way `--reverse-errors` would actually be usable by itself.

I'll open a quick PR adding this feature for the height cropping logic. Sorry I didn't think about this earlier! Then I think a new small realise would make sense!

ndmitchell:
I tend to release when I think there will be nothing interesting coming in the near future, or when the improvements are very worth having. It's not really time based.

Cool about reverse errors, I'll release after that.

drewboardman:
Hey this is a great update. Any idea when I'd be able to get it locally?

ndmitchell:
@drewboardman - it's available in ghcid 0.7.4 released yesterday.

drewboardman:
Nice! Forgive my lack of knowledge, but if I use stack for ghcid can I just run "stack upgrade"?

basile-henry:
I don't personally use stack, but I believe you can get it from the latest nightly release:

```sh
stack --resolver nightly-2019-04-18 install ghcid
```

drewboardman:
Will this be compatible with ghc 8.6.4?

ndmitchell:
@drewboardman ghcid is generally compatible with all versions of GHC - it's very loosely coupled. It's almost certainly compatible with 8.6.4, but if not, raise a ticket (I think I've even tried that specific version).

PR #238, opened 2019-04-17
Properly handle term height limit when reversing errors

basile-henry:
Quick fix to improve #237

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -268,12 +268,24 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
                     Nothing -> return (wrap msg, wrap load, [])
                     Just termHeight -> do
                         (termHeight, msg) <- return $ takeRemainder termHeight $ wrap msg
-                        (termHeight, load) <- return $ takeRemainder termHeight $ wrap load
+                        (termHeight, load) <-
+                            let takeRemainder' =
+                                    if reverse_errors
+                                    then -- When reversing the errors we want to crop out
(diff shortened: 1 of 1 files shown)


src/Ghcid.hs
```diff
@@ -268,12 +268,24 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
...
                         (termHeight, msg) <- return $ takeRemainder termHeight $ wrap msg
-                        (termHeight, load) <- return $ takeRemainder termHeight $ wrap load
+                        (termHeight, load) <-
+                            let takeRemainder' =
+                                    if reverse_errors
+                                    then -- When reversing the errors we want to crop out
+                                         -- the top instead of the bottom of the load
+                                         fmap reverse . takeRemainder termHeight . reverse
+                                    else takeRemainder termHeight
+                            in return $ takeRemainder' $ wrap load
```
basile-henry: At this point the load is more general than just error messages, is it not? If that's the case then this will break test outputs or other non errors/warning outputs.

I'm going to have to think a bit more about this _feature_. :thinking:

basile-henry: It looks like test outputs might not use the `outputFill` function to print their output, so maybe it's fine.

ndmitchell:
Thanks!

ndmitchell:
Release now?

basile-henry:
Yes! Whenever is best for you! :smile:
Thank you!

ndmitchell:
Released as 0.7.4

PR #240, opened 2019-04-21
Clarify how --run & --test are alike

dwijnand:
I found the mention of `--test` in the changelog to bring clarity to how `--run` and `--test` are one and the same outside of the default.  I propose adding that to the help message.  It adds to its length, but do you think it's a good addition?


ndmitchell:
If it helps you can't be a bad thing! Thanks for the patch

PR #245, opened 2019-05-09
ghcid: link with -rtsopts to be able to use RTS

domenkozar:
On our project, ghcid eats up 30G of memory
and freezes my laptop, so I'd like to limit that using 
RTSOPTS="-M4M" but ghcid is not compiled with
rtsopts support yet.

### DIFF
--- a/ghcid.cabal
+++ b/ghcid.cabal
@@ -47,7 +47,7 @@ library
 executable ghcid
     hs-source-dirs: src
     default-language: Haskell2010
-    ghc-options: -main-is Ghcid.main -threaded
+    ghc-options: -main-is Ghcid.main -threaded -rtsopts
     main-is: Ghcid.hs
     build-depends:
         base >= 4.7 && < 5,


ndmitchell:
Sure. But... There's nothing in `ghcid` that should get anywhere close to 1Gb even. `ghcid` doesn't load up any `GHC` code in it. For context, my `ghcid` instances on my machine at the moment (which have been up since this morning) are 8Mb and 10Mb. I'd be very keen if you can shed any insight into how it gets to that size. Does it get that big on startup? Or only after you've used it for a while? Does it creep up slowly? Does something specific cause it to jump?

`ghcid` should occupy memory linear in the number of active warnings/errors, plus the number of loaded files. I have made no attempt at minimising memory (e.g. I use `String`) because neither of those numbers are expected to be huge in any sane projects. If that's an incorrect assumption I can reduce memory once we know what is the culprit.

domenkozar:
@ndmitchell I think this is the unfortunate consequence that we'd like to set `RTSOPTS` enviroment variable for `--command stack repl`, but ghcid exists as it's not supported for that process.

Thank you!

ndmitchell:
@domenkozar not entirely understanding you... What is using the huge amount of memory? `ghcid` itself? Or a process `ghcid` spawns?

PR #248, opened 2019-05-31
Implement "Rethinking REPLs"

isovector:
This PR implements the [Unison notebook](http://unisonweb.org/2018-09-25/repl.html#rethinking-repls-and-notebooks) idea --- essentially adding support for interactive REPL commands directly in source code.

The idea is that I can write the following:

```haskell
magic :: String -> String
magic = reverse

-- $> magic "hello world"
```

to which, ghcid will now reply:

```
All good (25 modules, at 17:42:54)

/home/sandy/prj/polysemy/src/Polysemy.hs:160:1
$> magic "hello world"
"dlrow olleh"
```

In essence, this lets you run little repl commands directly in your source file.

Link to the original idea in Unison: https://github.com/unisonweb/unison/pull/252

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -258,7 +258,7 @@ runGhcid :: Session -> Waiter -> IO TermSize -> ([String] -> IO ()) -> Options -
 runGhcid session waiter termSize termOutput opts@Options{..} = do
     let limitMessages = maybe id (take . max 1) max_messages
 
-    let outputFill :: String -> Maybe (Int, [Load]) -> [((FilePath, (Int, Int)), (String, String))] -> [String] -> IO ()
+    let outputFill :: String -> Maybe (Int, [Load]) -> [EvalResult] -> [String] -> IO ()
         outputFill currTime load evals msg = do
             load <- return $ case load of
                 Nothing -> []
@@ -316,7 +316,7 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
                 outStrLn $ "%MESSAGES: " ++ show messages
                 outStrLn $ "%LOADED: " ++ show loaded
 
-            let evals = [((loadFile e, loadFilePos e), (evalCommand e, evalResult e)) | e@EvalResult {} <- messages]
+            let evals = [e | Eval e <- messages]
             let (countErrors, countWarnings) = both sum $ unzip
                     [if loadSeverity == Error then (1,0) else (0,1) | m@Message{..} <- messages, loadMessage /= []]
             let hasErrors = countErrors /= 0 || (countWarnings /= 0 && not warnings)
@@ -389,14 +389,14 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
 
 
 -- | Given an available height, and a set of messages to display, show them as best you can.
-prettyOutput :: String -> Int -> [Load] -> [((FilePath, (Int, Int)), (String, String))] -> [String]
+prettyOutput :: String -> Int -> [Load] -> [EvalResult] -> [String]
 prettyOutput currTime loadedCount [] evals =
     (allGoodMessage ++ " (" ++ show loadedCount ++ " module" ++ ['s' | loadedCount /= 1] ++ ", at " ++ currTime ++ ")")
         : concatMap printEval evals
 prettyOutput _ _ xs evals = concatMap loadMessage xs ++ concatMap printEval evals
 
-printEval :: ((FilePath, (Int, Int)), (String, String)) -> [String]
-printEval ((file, (line, col)), (msg, result)) =
+printEval :: EvalResult -> [String]
+printEval (EvalResult file (line, col) msg result) =
   [ " "
     , concat
         [ file
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -100,7 +100,7 @@ options = cmdArgsMode $ Options
     ,max_messages = Nothing &= name "n" &= help "Maximum number of messages to print"
     ,color = Auto &= name "colour" &= name "color" &= opt Always &= typ "always/never/auto" &= help "Color output (defaults to when the terminal supports it)"
     ,setup = [] &= name "setup" &= typ "COMMAND" &= help "Setup commands to pass to ghci on stdin, usually :set <something>"
-    ,allow_eval = False &= name "allow-eval" &= help "Allow execution of arbitrary REPL commands in comments"
+    ,allow_eval = False &= name "allow-eval" &= help "Execute REPL commands in comments"
     } &= verbosity &=
     program "ghcid" &= summary ("Auto reloading GHCi daemon v" ++ showVersion version)
 
@@ -398,14 +398,14 @@ prettyOutput _ _ xs evals = concatMap loadMessage xs ++ concatMap printEval eval
 printEval :: ((FilePath, (Int, Int)), (String, String)) -> [String]
 printEval ((file, (line, col)), (msg, result)) = do
   [ " "
-    , mconcat
+    , concat
         [ file
         , ":"
         , show line
         , ":"
         , show col
         ]
-    ] ++ (zipWith (++) (repeat "λ> ") $ lines msg)
+    ] ++ (map ("$> " ++) $ lines msg)
       ++ lines result
 
 
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -391,12 +391,12 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
 -- | Given an available height, and a set of messages to display, show them as best you can.
 prettyOutput :: String -> Int -> [Load] -> [((FilePath, (Int, Int)), (String, String))] -> [String]
 prettyOutput currTime loadedCount [] evals =
-    [allGoodMessage ++ " (" ++ show loadedCount ++ " module" ++ ['s' | loadedCount /= 1] ++ ", at " ++ currTime ++ ")"
-    ] ++ concatMap printEval evals
+    (allGoodMessage ++ " (" ++ show loadedCount ++ " module" ++ ['s' | loadedCount /= 1] ++ ", at " ++ currTime ++ ")")
+        : concatMap printEval evals
 prettyOutput _ _ xs evals = concatMap loadMessage xs ++ concatMap printEval evals
 
 printEval :: ((FilePath, (Int, Int)), (String, String)) -> [String]
-printEval ((file, (line, col)), (msg, result)) = do
+printEval ((file, (line, col)), (msg, result)) =
   [ " "
     , concat
         [ file
@@ -405,7 +405,7 @@ printEval ((file, (line, col)), (msg, result)) = do
         , ":"
         , show col
         ]
-    ] ++ (map ("$> " ++) $ lines msg)
+    ] ++ map ("$> " ++) (lines msg)
       ++ lines result
 
 
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -63,6 +63,7 @@ data Options = Options
     ,max_messages :: Maybe Int
     ,color :: ColorMode
     ,setup :: [String]
+    ,allow_eval :: Bool
     }
     deriving (Data,Typeable,Show)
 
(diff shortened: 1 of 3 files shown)


isovector:
I'm happy to clean up the lint if you like the idea. Also, I just realized that maybe `-- >` isn't a good choice, since I think it has meaning to haddock.

isovector:
@jkeuhlen made the good point that this can let people run arbitrary code. It should probably be hidden behind a flag :)

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
     sessionStart session cmd setup
 
 
+performEvals :: Ghci -> Bool -> [FilePath] -> IO [Load]
+performEvals _ False _ = pure []
+performEvals ghci True reloaded = do
+  cmds <- traverse getCommands reloaded
```
ndmitchell: The rest of the file is 4 space indentation so best to stick with that please.

ndmitchell: Is this just a mapM?

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
...
+  fmap join $ flip traverse cmds $ \(file, cmds') ->
```
ndmitchell: Is this traverse just a forM?

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
...
+    flip traverse cmds' $ \(num, cmd) -> do
```
ndmitchell: And this one too?

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
...
+      ref <- newIORef []
+      execStream ghci (intercalate " " $ lines cmd) $ \_ resp -> modifyIORef ref (++ resp)
```
ndmitchell: This is probably an _O(n^2)_ algorithm. I'd suggest `modifyIORef ref (resp:)` and then after the `readIORef` doing a `concat . reverse`

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
...
+      resp <- readIORef ref
+      pure $ EvalResult file (num, 1) cmd resp
+
+
+getCommands :: FilePath -> IO (FilePath, [(Int, String)])
+getCommands fp = do
+  ls <- readFile fp
```
ndmitchell: Should this be readFileUTF8'? I suspect so.

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
...
+  pure (fp, splitCommands $ zip [1..] $ lines ls)
```
ndmitchell: I tend to use `return` over `pure`. There is also `zipFrom 1` to add indicies to the lines.

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
...
+
+splitCommands :: [(Int, String)] -> [(Int, String)]
+splitCommands [] = []
+splitCommands ((num, line) : ls)
+  | isCommand line =
```
ndmitchell: See the stripPrefix function which can avoid the drop and the isPrefixOf pair.

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
...
+      let (cmds, xs) = span (isCommand . snd) ls
+       in (num, unlines $ fmap (drop $ length commandPrefix) $ line : fmap snd cmds) : splitCommands xs
```
ndmitchell: So N successive lines are treated as a single block, rather than as N successive tests? But then sent over as N lines?

src/Language/Haskell/Ghcid/Types.hs
```diff
@@ -45,6 +45,13 @@ data Load
       LoadConfig
         {loadFile :: FilePath -- ^ The file that was being loaded, @.ghci@.
         }
+    | -- | A config file was loaded, usually a .ghci file (GHC 8.2 and above only)
```
ndmitchell: Is this comment now wrong?

src/Session.hs
```diff
@@ -125,6 +134,40 @@ sessionRestart session@Session{..} = do
     sessionStart session cmd setup
 
 
+performEvals :: Ghci -> Bool -> [FilePath] -> IO [Load]
```
ndmitchell: I'm wondering if Load should contain PerformEvals or not? Should it return a separate type, and then higher level things work with Either Load EvalResult? The Session seems to mostly deal with evals (as seems entirely reasonable), but the Load data type is defined in the ghcid library, which means the data type appears there despite not being used there.

src/Session.hs
```diff
@@ -3,11 +3,12 @@
 -- | A persistent version of the Ghci session, encoding lots of semantics on top.
 --   Not suitable for calling multithreaded.
 module Session(
-    Session, withSession,
+    Session(allowEval), withSession,
```
ndmitchell: I find the approach of exposing one accessor to Session a bit unpleasant, partly because you can then mutate it later as you wish. Maybe I'm fussing though - it seems to work out relatively nicely below...

ndmitchell: I think my preference would be an `enableEval :: Session -> Session` though.

src/Ghcid.hs
```diff
@@ -375,21 +378,35 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
...
+printEval ((file, (line, col)), (msg, result)) = do
+  [ " "
+    , mconcat
+        [ file
+        , ":"
+        , show line
+        , ":"
+        , show col
+        ]
+    ] ++ (zipWith (++) (repeat "λ> ") $ lines msg)
```
ndmitchell: Normally I try and avoid Unicode since otherwise you are causing issues for a set of people who would otherwise have had their code work. Why not make this `$>` instead?

ndmitchell: `map ("$> " ++) $ lines msg`

src/Ghcid.hs
```diff
@@ -375,21 +378,35 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
...
+prettyOutput :: String -> Int -> [Load] -> [((FilePath, (Int, Int)), (String, String))] -> [String]
+prettyOutput currTime loadedCount [] evals =
+    [allGoodMessage ++ " (" ++ show loadedCount ++ " module" ++ ['s' | loadedCount /= 1] ++ ", at " ++ currTime ++ ")"
+    ] ++ concatMap printEval evals
+prettyOutput _ _ xs evals = concatMap loadMessage xs ++ concatMap printEval evals
+
+printEval :: ((FilePath, (Int, Int)), (String, String)) -> [String]
+printEval ((file, (line, col)), (msg, result)) = do
+  [ " "
+    , mconcat
```
ndmitchell: I'd just use concat here.

src/Ghcid.hs
```diff
@@ -99,6 +100,7 @@ options = cmdArgsMode $ Options
     ,max_messages = Nothing &= name "n" &= help "Maximum number of messages to print"
     ,color = Auto &= name "colour" &= name "color" &= opt Always &= typ "always/never/auto" &= help "Color output (defaults to when the terminal supports it)"
     ,setup = [] &= name "setup" &= typ "COMMAND" &= help "Setup commands to pass to ghci on stdin, usually :set <something>"
+    ,allow_eval = False &= name "allow-eval" &= help "Allow execution of arbitrary REPL commands in comments"
```
ndmitchell:
```suggestion
    ,allow_eval = False &= name "allow-eval" &= help "Execute REPL commands in comments"
```

src/Ghcid.hs
```diff
@@ -256,11 +258,11 @@ runGhcid :: Session -> Waiter -> IO TermSize -> ([String] -> IO ()) -> Options -
 runGhcid session waiter termSize termOutput opts@Options{..} = do
     let limitMessages = maybe id (take . max 1) max_messages
 
-    let outputFill :: String -> Maybe (Int, [Load]) -> [String] -> IO ()
-        outputFill currTime load msg = do
+    let outputFill :: String -> Maybe (Int, [Load]) -> [((FilePath, (Int, Int)), (String, String))] -> [String] -> IO ()
```
ndmitchell: If we had a separate `EvalResult` type (which seems like a good idea anyway) then this bit type argument would have that type, which would be good.

isovector:
Excited to hear that you like the idea! In addition to (or instead of) the `--allow-eval` flag, I was thinking  we should just never run any commands that are there _before you start ghcid._  That way nobody can drop malicious snippets into a repo somewhere. What do you think?

I'll clean up the lint today.

src/Session.hs
isovector: Yeah, I wasn't in love with it either, but the thought of threading `allow_eval` through every function seemed less pleasant. I'm happy to put in the work if you'd prefer it -- maybe as a `ReaderT` so anyone else who might want to do so in the future will have less work?

src/Session.hs
isovector: I don't have enough context of how ghcid works to have an opinion here. I didn't even realize there was a library vs an app, but it would explain why `stack repl` refuses to load the project! Happy to do whatever you suggest, though this seemed like the most straightforward impl.

src/Session.hs
isovector: Yeah, so if we write this:

```haskell
-- > let x = 5
-- >  in x
-- 
-- > let y = 6
-- >  in y
```

this will show up as two commands. Ideally we'd like to preserve the newlines, and then dump them into ghci via `:{ multiline blocks :}`, but ghcid doesn't parse ghci's response to such things correctly. Keeping the newline structure here paves the way for getting multiline working properly in the future, though today we just glom the lines together with spaces and hope it works.

src/Session.hs
isovector: Cool, I didn't know about this! I'm not entirely sure how to make the change though, since we still want to `span` over the next lines.

src/Session.hs
ndmitchell: I was thinking of keeping allowEval in Session, but not exposing the allowEval (which gives you a lot of power), and instead just `enableEval s = s{allowEval=True}`. That way `Session` remains entirely abstract, but you can keep the rest of the code the same, without any ReaderT or similar.

isovector:
Sounds great! I'll try to get this polished up tomorrow.

isovector:
@ndmitchell I think I've responded to all of your suggestions. CI is broken, but it appears to be for unrelated reasons:

```
Using compiler: ghc-8.6.5
Asking cabal to calculate a build plan...
Trying with packages from nightly-2019-06-06 as hard constraints...
The following lines from cabal-install output could not be parsed: 
ansi-terminal-0.9.1 (via: ghcid-0.7.4 ghcid-0.7.4 ghcid-0.7.4 tasty-1.2.2
ansi-wl-pprint-0.6.9) (new package)
CallStack (from HasCallStack):
  error, called at src/Stack\Solver.hs:174:16 in stack-1.9.3-3S9JDRpC5h6C7YOiQDGqyN:Stack.Solver
```

ethercrow:
@isovector thank you for this feature! Just tried it and had only one minor inconvenience with language pragmas not affecting the ghcid repl.

```
{-# LANGUAGE TypeApplications #-}

-- $> map (succ @Int) [1, 2]
```

results in

```
$> map (succ @Int) [1, 2]
<interactive>:59:6-14: error:    Pattern syntax in expression context: succ@Int    Did you mean to
enable TypeApplications?
```

but is fine if I add 

```
-- $> :set -XTypeApplications
```

I guess it could be fixed by not only collecting lines starting with `-- $>` but also converting `{-# language blah #-}` to `:set -Xblah`.

isovector:
@ethercrow yeah, I've been running into that too. There's also an issue with the imports being in the wrong scope. Perhaps @ndmitchell has some ideas about doing this more correctly.

isovector:
Gentle ping on this.

ndmitchell:
Sorry, Zurihac burnt out all my brain cycles for a few days and I'm still in catch up. I'll do a real review tomorrow (when my brain has returned)

ndmitchell:
Thanks, took a few more days for my brain to recover, but it looks all good. Could I trouble you to write a paragraph or so for the README, explaining what the feature is and roughly how to use it? Otherwise its unlikely to be discoverable for users.

fosskers:
`ghcid --allow-eval` works like a charm!

PR #262, opened 2019-07-01
Add --test-message flag

### LINKED ISSUE #260: Add a way to customize the shown `--test` message
Right now when you run `ghcid` with `--test=...` it shows:

```
Running test...
```

just before the "test" is run.

In our case we are starting a web server in `--test`, so seeing "Running test..." just before bringing up the web server (which might take a while) is a bit strange. Also, for developers joining our project and running our shake commands we have to explain them "This is not actually running any test".

Do you think it would be a good idea to customize this message? It could be something like `--test-message="Starting web server..."`. I know it might feel a bit redundant to pass a message to be shown right next, but it makes sense if the command runs as part of other commands.

asterite:
Fixes #260

The --test-message flag allows customizing the message shown by the --test flag, which defaults to "Running test...".

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -42,6 +42,7 @@ data Options = Options
     {command :: String
     ,arguments :: [String]
     ,test :: [String]
+    ,test_message :: String
     ,run :: [String]
     ,warnings :: Bool
     ,lint :: Maybe String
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks - that's even simpler than I thought it would be! Will merge once the CI finishes.

ndmitchell:
Thanks!

asterite:
@ndmitchell Thank you! The code was really easy to get into.

PR #281, opened 2019-10-04
Improvements for emacs

yoricksijsling:
I found the emacs mode quite buggy. These are the changes I've been using locally for a while now.

The window creation behaviour has changed, you can recover the old behaviour by putting the
following in your init.el:

    (add-to-list
     'display-buffer-alist
     '(\"*ghcid*\"
       (display-buffer-reuse-window   ;; First try to reuse an existing window
        display-buffer-at-bottom      ;; Then try a new window at the bottom
        display-buffer-pop-up-window) ;; Otherwise show a pop-up
       (window-height . 18)      ;; New window will be 18 lines
       ))

Improvements:

* Don't assume that nlinum-mode is available.
* Respect user configuration when displaying the buffer. Now it works with purpose-mode and custom
  `display-buffer-alist`s
* Automatically detect height of ghcid window.
* Use the standard `with-selected-window` to temporarily select the ghcid window.
* Use term-line-mode instead of term-char-mode, so the usual navigation shortcuts work in the ghcid
  buffer.
* Resets the directory of the ghcid process if you call `ghcid` again.
* Fix bug where the `next-error` would not always go the the right place in a file.

Thanks for the pull request!

### DIFF
--- a/plugins/emacs/ghcid.el
+++ b/plugins/emacs/ghcid.el
@@ -1,74 +1,133 @@
 ;; Really basic ghcid+stack support in emacs with compilation-mode
 ;; Use M-x ghcid to launch
 
+(require 'term)
+
 ;; Set ghcid-target to change the stack target
 (setq ghcid-target "")
 
-(setq ghcid-height 15)
-(defun ghcid-stack-cmd (target)
-      (format "stack ghci %s --test --bench --ghci-options=-fno-code" target))
+(setq ghcid-process-name "ghcid")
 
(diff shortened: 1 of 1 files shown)


ndmitchell:
@WraithM - can you review? Totally out of my depth here.

wraithm:
@ndmitchell This is great! It works so much better than my version.

Thanks @yoricksijsling. Well done!

ndmitchell:
Thanks!

PR #285, opened 2019-10-31
Preserve newlines in eval output

pbevin:
In `--allow-eval`, newlines are removed from the output. This fixes it to preserve line breaks.


ndmitchell:
Seems plausible - thanks!

PR #287, opened 2019-11-13
Only lint reloaded files

alexfmpe:


ndmitchell:
Very nice! Thanks. I'll make a release with this change, but probably not for a few weeks

ndmitchell:
Released in 0.7.7

PR #288, opened 2019-11-29
Handle GHCI unexpected errors better

matthewbauer:
First reported in
https://github.com/obsidiansystems/obelisk/issues/563, we see an issue
where ghci crashes had bad error messages in ghcid. Only information
given was "unexpected error" and the command that was run.

To better handle this, we need to modify the “consume” functions to
better keep track of error messages. This changes the type signature
from a Maybe to an Either:

```diff
-consume :: Stream -> (String -> IO (Maybe a)) -> IO (Maybe a)
+consume :: Stream -> (String -> IO (Maybe a)) -> IO (Either (Maybe String) a)
```

This allows us to keep track of what the last message from GHCI was so
so that we can report it back to the user. Similarly, the signature
for GhciError was changed:

```diff
-data GhciError = UnexpectedExit String String
+data GhciError = UnexpectedExit String String (Maybe String)
```

So that we can store this information. After that, just change the
error handling to print a Just "..." message.

ghcid now shows something like this on out of memory:

  Command "ghci ..." exited unexpectedly with error message: ghc:
  failed to create OS thread: Cannot allocate memory

Thanks for the pull request!

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -181,10 +181,20 @@ data TermSize = TermSize
     ,termWrap :: WordWrap
     }
 
+-- | Modify IO to catch runtime exceptions.
+handleErrors :: IO () -> IO ()
+handleErrors = handle $ \err -> case err of
+    UnexpectedExit cmd _ mmsg -> do
+        putStr $ "Command \"" ++ cmd ++ "\" exited unexpectedly"
+        case mmsg of
+            Just msg -> putStrLn $ " with error message: " <> msg
+            Nothing -> putStrLn ""
+        exitFailure
+
 -- | Like 'main', but run with a fake terminal for testing
 mainWithTerminal :: IO TermSize -> ([String] -> IO ()) -> IO ()
 mainWithTerminal termSize termOutput =
-    handle (\(UnexpectedExit cmd _) -> do putStrLn $ "Command \"" ++ cmd ++ "\" exited unexpectedly"; exitFailure) $
+    handleErrors $
         forever $ withWindowIcon $ withSession $ \session -> do
             setVerbosity Normal -- undo any --verbose flags
 
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -181,15 +181,14 @@ data TermSize = TermSize
     ,termWrap :: WordWrap
     }
 
--- | Modify IO to catch runtime exceptions.
+-- | On the 'UnexpectedExit' exception exit with a nice error message.
 handleErrors :: IO () -> IO ()
-handleErrors = handle $ \err -> case err of
-    UnexpectedExit cmd _ mmsg -> do
-        putStr $ "Command \"" ++ cmd ++ "\" exited unexpectedly"
-        case mmsg of
-            Just msg -> putStrLn $ " with error message: " <> msg
-            Nothing -> putStrLn ""
-        exitFailure
+handleErrors = handle $ \(UnexpectedExit cmd _ mmsg) -> do
+    putStr $ "Command \"" ++ cmd ++ "\" exited unexpectedly"
+    putStrLn $ case mmsg of
+        Just msg -> " with error message: " <> msg
+        Nothing -> ""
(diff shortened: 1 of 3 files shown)


src/Ghcid.hs
```diff
@@ -181,10 +181,20 @@ data TermSize = TermSize
     ,termWrap :: WordWrap
     }
 
+-- | Modify IO to catch runtime exceptions.
+handleErrors :: IO () -> IO ()
+handleErrors = handle $ \err -> case err of
```
ndmitchell: This `case` seems redundant - why not just `\(UnexpectedExist cmd _ mmsg) ->`. FWIW, all the CI failures are because HLint things a LambdaCase would be better here - but actually the lambda case isn't required.

src/Ghcid.hs
```diff
@@ -181,10 +181,20 @@ data TermSize = TermSize
     ,termWrap :: WordWrap
     }
 
+-- | Modify IO to catch runtime exceptions.
```
ndmitchell: I find this comment a bit misleading. Why not:
```suggestion
-- | On the 'UnexpectedExit' exception exit with a nice error message.
```

src/Language/Haskell/Ghcid.hs
```diff
@@ -118,11 +119,14 @@ startGhciProcess process echo0 = do
...
                 res1 <- res1
                 res2 <- res2
-                case liftM2 (,) res1 res2 of
-                    Nothing -> case cmdspec process of
-                        ShellCommand cmd -> throwIO $ UnexpectedExit cmd msg
-                        RawCommand exe args -> throwIO $ UnexpectedExit (unwords (exe:args)) msg
-                    Just v -> return v
+                case (res1, res2) of
+                    (Right v1, Right v2) -> return (v1, v2)
+                    (_, Left err) -> case cmdspec process of
```
ndmitchell: This bit is a tag duplicative. Can we `let raise msg err = throwIO $ case cmdspec process of ShellCommand cmd -> UnexpectedExit cmd msg err; ...` or similar? Anything that stops the logic being repeated twice.

src/Ghcid.hs
```diff
@@ -181,10 +181,20 @@ data TermSize = TermSize
...
+handleErrors :: IO () -> IO ()
+handleErrors = handle $ \err -> case err of
+    UnexpectedExit cmd _ mmsg -> do
+        putStr $ "Command \"" ++ cmd ++ "\" exited unexpectedly"
+        case mmsg of
+            Just msg -> putStrLn $ " with error message: " <> msg
+            Nothing -> putStrLn ""
```
3noch: Can factor out `putStrLn` here too.

src/Language/Haskell/Ghcid/Types.hs
```diff
@@ -12,7 +12,7 @@ import Data.Data
 import Control.Exception.Base (Exception)
 
 -- | GHCi shut down
-data GhciError = UnexpectedExit String String
+data GhciError = UnexpectedExit String String (Maybe String)
```
3noch: This feels record-syntax-worthy...

src/Language/Haskell/Ghcid/Types.hs
```diff
@@ -12,8 +12,11 @@ import Data.Data
 import Control.Exception.Base (Exception)
 
 -- | GHCi shut down
-data GhciError = UnexpectedExit String String
-    deriving (Show,Eq,Ord,Typeable,Data)
+data GhciError = UnexpectedExit {
+    _ghciErrorCmd :: String
```
ndmitchell: Why the leading underscores? These types have no lenses (which is the common reason for doing that), but I'd rather they were just unadorned non-underscore. I'll merge and fix up after.

ndmitchell:
Thanks! All merged. I'll make a release with this once the CI passes.

ndmitchell:
Released as 0.8

matthewbauer:
Awesome!

Issue #273, opened 2019-08-16
Passing a directory to --restart behaves like passing the directory to --reload


nfrisby:
My OS: Debian 9.

When I change a file under a directory passed to `--restart`, `ghcid` should restart but instead only reloads. Some details below.

#56 allows `--restart` to accept directories, by expanding any directories "just in time" during the check. Note how `files` is locally redefined by expanding any directory names in it.

https://github.com/ndmitchell/ghcid/blob/17775959407125814f782ba39751112e8c956ed9/src/Wait.hs#L63-L64

However, the logic in the main loop only restarts if the _original argument_ to `--restart` has a newer modification time. Otherwise it just reloads. On my OS, changing a file's mtime does not change all of its ancestory directories' mtime -- that's why the logic currently on `master` is misbehaving in this case.

https://github.com/ndmitchell/ghcid/blob/17775959407125814f782ba39751112e8c956ed9/src/Ghcid.hs#L372-L375

PR #300, opened 2020-02-12
Restart on file changes detected by --restart=dir/

martijnbastiaan:
This attempts to fix https://github.com/ndmitchell/ghcid/issues/273 by detecting whether any of the triggered paths are subfiles of a directory supplied `--restart`. As a consequence, it would override more specific paths passed to reload. E.g. in:

```
ghcid --reload=dir/path/specific.hs --restart=dir/path
```

`restart` would take precedence over `reload` if `dir/path/specific.hs` changes.

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -263,7 +263,7 @@ data ReloadMode = Reload | Restart deriving (Show, Ord, Eq)
 
 -- If we return successfully, we restart the whole process
 -- Use Continue not () so that inadvertant exits don't restart
-runGhcid :: Session -> Waiter ReloadMode -> IO TermSize -> ([String] -> IO ()) -> Options -> IO Continue
+runGhcid :: Session -> Waiter -> IO TermSize -> ([String] -> IO ()) -> Options -> IO Continue
 runGhcid session waiter termSize termOutput opts@Options{..} = do
     let limitMessages = maybe id (take . max 1) max_messages
 
@@ -394,6 +394,8 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
                     (Reload, ["Error when waiting, if this happens repeatedly, raise a ghcid bug.", err])
                   Right files ->
                     case partition (\(f, mode) -> mode == Reload) files of
+                      -- Prefer restarts over reloads. E.g., in case of both '--reload=dir'
+                      -- and '--restart=dir', ghcid would restart instead of reload.
                       (_, rs@(_:_)) -> (Restart, map fst rs)
                       (rl, _) -> (Reload, map fst rl)
 
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -259,9 +259,11 @@ main = mainWithTerminal termSize termOutput
 
 data Continue = Continue
 
+data ReloadMode = Reload | Restart deriving (Show, Ord, Eq)
+
 -- If we return successfully, we restart the whole process
 -- Use Continue not () so that inadvertant exits don't restart
-runGhcid :: Session -> Waiter -> IO TermSize -> ([String] -> IO ()) -> Options -> IO Continue
+runGhcid :: Session -> Waiter ReloadMode -> IO TermSize -> ([String] -> IO ()) -> Options -> IO Continue
 runGhcid session waiter termSize termOutput opts@Options{..} = do
     let limitMessages = maybe id (take . max 1) max_messages
 
@@ -316,7 +318,12 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
     project <- if project /= "" then return project else takeFileName <$> getCurrentDirectory
 
     -- fire, given a waiter, the messages/loaded/touched
-    let fire nextWait (messages, loaded, touched) = do
+    let
+      fire
+        :: ([(FilePath, ReloadMode)] -> IO (Either String [(FilePath, ReloadMode)]))
+        -> ([Load], [FilePath], [FilePath])
+        -> IO Continue
+      fire nextWait (messages, loaded, touched) = do
             currTime <- getShortTime
             let loadedCount = length loaded
             whenLoud $ do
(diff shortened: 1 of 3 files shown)


src/Ghcid.hs
```diff
@@ -308,7 +308,10 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
         putStrLn $ "\nNo files loaded, GHCi is not working properly.\nCommand: " ++ command
         exitFailure
 
-    restart <- return $ nubOrd $ restart ++ [x | LoadConfig x <- messages]
+    let stripTrailingSep = reverse . dropWhile (`elem` pathSeparators) . reverse
```
ndmitchell: This is just https://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:dropTrailingPathSeparator ?

src/Ghcid.hs
```diff
@@ -308,7 +308,10 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
...
+    curDir <- getCurrentDirectory
+    let absRestart = map (stripTrailingSep . (curDir </>)) restart
```
ndmitchell: Is this whole purpose of this bit just https://hackage.haskell.org/package/directory-1.3.6.0/docs/System-Directory.html#v:makeAbsolute ?

src/Ghcid.hs
```diff
@@ -382,8 +385,9 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
             whenLoud $ outStrLn $ "%RELOADING: " ++ unwords reason
             restartTimes2 <- mapM getModTime restart
             let restartChanged = [s | (False, s) <- zip (zipWith (==) restartTimes restartTimes2) restart]
+                restartFilesChanged = or [(ra ++ [pathSeparator]) `isPrefixOf` rb | ra <- absRestart, rb <- reason]
```
ndmitchell: See `addTrailingPathSeparator`.

ndmitchell: This line is pretty subtle - can you include a link to https://github.com/ndmitchell/ghcid/issues/273 so the reasoning is clear.

martijnbastiaan:
Yeah, I agree, the whole patch wasn't really satisfying to write and felt pretty hacky. I didn't realize I could use the dependencies / functions you linked, I'll update the patch tomorrow.

ndmitchell:
Thanks! Thinking about how it should be done "properly" I think the answer is to change:

```
waitFiles :: Waiter -> IO ([FilePath] -> IO [String])
```

To something like:

```
waitFiles :: Waiter -> IO ([(FilePath, a)] -> IO [(String, a)])
```

And then raise all the `a` values that match. You would use `data Mode = Restart | Reload` and then skip this whole checking the modtimes of restart files.

ndmitchell:
That said, it's a fair amount more work, so I'm happy to merge this approach with the little tweaks I suggested.

martijnbastiaan:
@ndmitchell I figured I'd do a refactor. I've added some type signatures here and there to help GHC give me sensible error messages. Could you do another review?

I'll look into the CI failures tomorrow.

src/Wait.hs
```diff
@@ -53,23 +55,36 @@ listContentsInside test dir = do
...
+    onError e = sleep 1.0 >> pure (Left (show e))
+
+    go :: UTCTime -> [(FilePath, a)] -> IO (Either String [(FilePath, a)])
+    go base files = do
+        whenLoud $ outStrLn $ "%WAITING: " ++ unwords (map fst files)
         -- As listContentsInside returns directories, we are waiting on them explicitly and so
         -- will pick up new files, as creating a new file changes the containing directory's modtime.
-        files <- fmap concat $ forM files $ \file ->
-            ifM (doesDirectoryExist file) (listContentsInside (return . not . isPrefixOf "." . takeFileName) file) (return [file])
+        files <- fmap concat $ forM files $ \(file, a) ->
```
ndmitchell: Use `concatForM`.

src/Wait.hs
```diff
@@ -53,23 +55,36 @@ listContentsInside test dir = do
...
+              (fmap (,a) <$> listContentsInside (return . not . isPrefixOf "." . takeFileName) file)
+              (return [(file, a)])
         case waiter of
             WaiterPoll t -> return ()
             WaiterNotify manager kick mp -> do
-                dirs <- fmap Set.fromList $ mapM canonicalizePathSafe $ nubOrd $ map takeDirectory files
+                dirs <-
+                  fmap Map.fromList
+                    $ mapM canonicalizePathSafe
+                    $ nubOrdOn fst
```
ndmitchell: This looks like a bug. If you have `--restart=foo --reload=foo` then the `nubOrdOn` is going to throw away one of them, whichever is second. If that's the invariant, it needs to be documented. If you think that isn't a problem cause we recover it later, then why do you need the Map in the first place?

martijnbastiaan: `fire` already explicitly lets restarts have priority over reloads, so I think it'd be neater to just not throw it away here at all.

martijnbastiaan: Thanks for catching!

src/Ghcid.hs
```diff
@@ -378,19 +385,32 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
...
-            whenLoud $ outStrLn $ "%RELOADING: " ++ unwords reason
-            restartTimes2 <- mapM getModTime restart
-            let restartChanged = [s | (False, s) <- zip (zipWith (==) restartTimes restartTimes2) restart]
+            reason <-
+              nextWait $
+                map (,Restart) restart ++
+                map (,Reload) reload ++
+                map (,Reload) loaded
+
+            let
```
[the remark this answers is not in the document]
martijnbastiaan: I was thinking of 'course of action', as it used to be either `Reload` or `Restart`. I added the reason/paths later. But, I guess it doesn't really work if you leave out the last two words anyway, so I'll rename it. Thanks!

martijnbastiaan:
Thanks for the comments! I'm going to call it a day now. I hope to fix the remaining issues this weekend.

martijnbastiaan:
@ndmitchell Should be good to go now!

src/Wait.hs
```diff
@@ -20,18 +22,22 @@ import System.FSNotify
...
 
-withWaiterNotify :: (Waiter -> IO a) -> IO a
+withWaiterNotify :: (Waiter a -> IO b) -> IO b
 withWaiterNotify f = withManagerConf defaultConfig{confDebounce=NoDebounce} $ \manager -> do
     mvar <- newEmptyMVar
     var <- newVar Map.empty
     f $ WaiterNotify manager mvar var
 
+firstM :: Monad m => (a -> m c) -> (a, b) -> m (c, b)
+firstM f (x, y) = fmap (, y) (f x)
```
martijnbastiaan: Wasn't sure about adding this one. It exists in other libraries, but I didn't want to add another dependency to `ghcid`.

ndmitchell: I just added it to extra-1.6.20 so if you adjust the dependency bounds you can use it from there.

ndmitchell:
Thanks for the adjustments. I'll review, hopefully later today, and if not, tomorrow.

src/Wait.hs
ndmitchell: It's now in extra-1.6.20 so if you change the .cabal to require it, then you can use it from there.

martijnbastiaan: Done!

martijnbastiaan:
Cool :). I've applied your suggestions once again.

src/Wait.hs
```diff
@@ -53,57 +56,65 @@ listContentsInside test dir = do
...
                 [] -> recheck files new
                 xs -> do
                     let disappeared = [x | (x, Just _, Nothing) <- zip3 files old new]
-                    when (disappeared /= []) $ do
+                    unless (null disappeared) $ do
                         -- if someone is deleting a needed file, give them some space to put the file back
                         -- typically caused by VIM
                         -- but try not to
-                        whenLoud $ outStrLn $ "%WAITING: Waiting max of 1s due to file removal, " ++ unwords disappeared
+                        whenLoud $ outStrLn $ "%WAITING: Waiting max of 1s due to file removal, " ++ unwords (map fst disappeared)
```
ndmitchell: Does this need a `nubOrd`, since if the file is being waited for twice, as Reload and Restart, it might appear twice in the output?

src/Wait.hs
```diff
@@ -53,57 +56,65 @@ listContentsInside test dir = do
 --   starting from when 'waitFiles' was initially called.
 --
 --   Returns a message about why you are continuing (usually a file name).
-waitFiles :: Waiter -> IO ([FilePath] -> IO [String])
+waitFiles :: forall a.  Ord a => Waiter a -> IO ([(FilePath, a)] -> IO (Either String [(FilePath, a)]))
```
ndmitchell: The use of Either here is much nicer 👍

src/Wait.hs
```diff
@@ -20,13 +22,14 @@ import System.FSNotify
 import Language.Haskell.Ghcid.Util
 
 
-data Waiter = WaiterPoll Seconds
-            | WaiterNotify WatchManager (MVar ()) (Var (Map.Map FilePath StopListening))
+data Waiter a
+  = WaiterPoll Seconds
+  | WaiterNotify WatchManager (MVar ()) (Var (Map.Map (FilePath, a) StopListening))
```
ndmitchell: Do we need the `a` in the `Map` at all? The WaiterNotify just records which watches we have out with the underlying inotify watcher. We only use the WaiterNotify to kick us if something happens, therefore I think removing the `a` here is safe, since we use the `recheck` loop to raise the right things. Does that make sense?

src/Wait.hs
martijnbastiaan: Good point! Fixed!

src/Wait.hs
martijnbastiaan: You're right, we don't! I've removed it.

martijnbastiaan:
Sorry for the wait @ndmitchell, I've applied your suggestions. Thanks again :)

ndmitchell:
Thanks a lot, and for all the revisions!

PR #302, opened 2020-02-20
Add --target option for specifying argument to `cabal repl`

Gabriel439:
90% of the time, when I want to customize `ghcid`'s behavior, it's to specify
an explicit target to `cabal repl`.  Usually I do so by providing
`--command 'cabal repl lib:foo'`, but we can simplify this to
`--target lib:foo` and also take advantage of other useful default
behaviors (like `--ghc-options=-fno-code`)

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -8,7 +8,6 @@ import Control.Exception
 import System.IO.Error
 import Control.Applicative
 import Control.Monad.Extra
-import qualified Data.Foldable as Foldable
 import Data.List.Extra
 import Data.Maybe
 import Data.Ord
@@ -161,8 +160,7 @@ autoOptions o@Options{..}
               | ".ghci" `elem` files -> f ("ghci":opts) [curdir </> ".ghci"]
               | cabal /= [] ->
                   let useCabal =
-                          let prefix =
-                                  [ "cabal", "repl" ] ++ Foldable.toList target
+                          let prefix = [ "cabal", "repl" ] ++ maybeToList target
                               suffix = map ("--ghc-options=" ++) opts
                           in  prefix ++ suffix
                       useGHCI = "cabal exec -- ghci":opts
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -8,6 +8,7 @@ import Control.Exception
 import System.IO.Error
 import Control.Applicative
 import Control.Monad.Extra
+import qualified Data.Foldable as Foldable
 import Data.List.Extra
 import Data.Maybe
 import Data.Ord
@@ -65,6 +66,7 @@ data Options = Options
     ,color :: ColorMode
     ,setup :: [String]
     ,allow_eval :: Bool
+    ,target :: Maybe String
     }
     deriving (Data,Typeable,Show)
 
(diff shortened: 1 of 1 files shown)


src/Ghcid.hs
```diff
@@ -156,7 +159,18 @@ autoOptions o@Options{..}
...
+                          let prefix =
+                                  [ "cabal", "repl" ] ++ Foldable.toList target
+
+                              suffix = map ("--ghc-options=" ++) opts
+
+                          in  prefix ++ suffix
+
+                      useGHCI = "cabal exec -- ghci":opts
+
+                  in  f (if null arguments then useCabal else useGHCI) cabal
```
ndmitchell: This section of the code now has a lot of vertical space. The blank lines between each line aren't needed. I do agree it was light on the whitespace before though!

src/Ghcid.hs
```diff
@@ -156,7 +159,18 @@ autoOptions o@Options{..}
                                 "stack exec --test --bench -- ghci" : opts
                 in f flags $ stack:cabal
               | ".ghci" `elem` files -> f ("ghci":opts) [curdir </> ".ghci"]
-              | cabal /= [] -> f (if null arguments then "cabal repl":map ("--ghc-options=" ++) opts else "cabal exec -- ghci":opts) cabal
+              | cabal /= [] ->
+                  let useCabal =
+                          let prefix =
+                                  [ "cabal", "repl" ] ++ Foldable.toList target
```
ndmitchell: I'd put this on one one. Better to use `maybeToList` rather than the more general Foldable.

src/Ghcid.hs
```diff
@@ -156,7 +159,18 @@ autoOptions o@Options{..}
...
+
+                              suffix = map ("--ghc-options=" ++) opts
+
+                          in  prefix ++ suffix
```
ndmitchell: I'd probably just write `let useCabal = ["cabal","repl"] ++ maybeToList target ++ map ("--ghc-options=" ++) opts`

ndmitchell:
Thanks! After this, I wonder if any arguments which match as a target should be automatically treated as target and follow the same logic?

basile-henry:
This looks like a great feature! Unfortunately I seem to be getting issues:

```
Warning: the following files would be used as linker inputs, but linking is not being done:
...
ghc: no input files
...
Command "cabal repl exe:project --ghc-options=-fno-code --ghc-options=-fno-break-on-exception --ghc-options=-fno-break-on-error --ghc-options=-v1 --ghc-options=-ferror-spans --ghc-options=-j" exited unexpectedly with error message:
```

By tweaking the command by hand it looks like the issue is `--ghc-options=-fno-code`. I guess it makes sense since my executable depends on the library in the same project.

I get this error consistently over several projects. Am I missing something? This seemed to be one of the selling points of this feature:

>  also take advantage of other useful default behaviors (like --ghc-options=-fno-code)

Maybe there are some other settings I can use to make this work?

Gabriel439:
@basile-henry: I probably missed that because I most commonly use `ghcid` on the `lib` target for a project (e.g. `--target lib:dhall`)

PR #308, opened 2020-03-13
Add convetional headers for Emacs package

vyorkin:
This PR adds a few headers for ghcid Emacs package which are required when using tools like [quelpa](https://github.com/quelpa/quelpa). For example, currently if you try to use `quelpa` + `use-package` to load the `ghcid` plugin:
```elisp
(use-package ghcid
  :after (haskell-mode)
  :quelpa
  (ghcid :fetcher url :url "https://raw.githubusercontent.com/ndmitchell/ghcid/master/plugins/emacs/ghcid.el")
  ...
  )
```
it will complain about missing headers:
```sh
Error getting PACKAGE-DESC: (error Package lacks a file header)
```
This PR provides a fix according to
https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html

(I hope that dates and copyrights are correct)

### DIFF
--- a/plugins/emacs/ghcid.el
+++ b/plugins/emacs/ghcid.el
@@ -1,9 +1,13 @@
 ;;; ghcid.el --- Really basic ghcid+stack support in emacs with compilation-mode -*- lexical-binding: t -*-
 
-;; Author: Neil Mitchell <ndmitchell@gmail.com>
-;; Maintainer: Neil Mitchell <ndmitchell@gmail.com>
+;; Author: Matthew Wraith <wraithm@gmail.com>
+;;         Yorick Sijsling
+;; Maintainer: Matthew Wraith <wraithm@gmail.com>
+;;             Yorick Sijsling
+;;             Vasiliy Yorkin <vasiliy.yorkin@gmail.com>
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks. Not too fussed on the date. For version, I'd be tempted to say 1.0 or similar and version the emacs stuff separate from ghcid - they aren't really tied to each other. For maintainer and author, I am neither for this .el file. Happy to put your name, or look at the history of this on GitHub. However, if there's no one better, I'm ok with being the placeholder name for both.

vyorkin:
Got it, thanks, I've updated the names and version

plugins/emacs/ghcid.el
```diff
@@ -98,7 +114,7 @@ exactly. See `ghcid-mode'."
 
       (term-exec (ghcid-buffer-name)
            ghcid-process-name
-           "/bin/bash"
+           "/run/current-system/sw/bin/bash"
```
ndmitchell: Is this a deliberate change?

vyorkin: Oh, sorry, this is added by accident :)

ndmitchell:
Thanks!

PR #311, opened 2020-03-16
`--target` use with Stack

### LINKED ISSUE #297: Component Selection
Hi there, I'm a working Haskeller and use `ghcid` every day. It really improves my development experience, so I can't thank you enough for it.

`ghcid` can get confused when loading projects that have more than one `Main` module, say from having multiple `executable` sections, or even just a single exec and single `test-suite`. To get around this, I do:
```
ghcid --command "stack ghci --ghci-options=-fno-code --ghci-options=-fno-break-on-exception --ghci-options=-fno-break-on-error --ghci-options=-v1 --ghci-options=-ferror-spans --ghci-options=-j -- foo:lib foo:test:foo-tests"
```
The plethora of options I had to borrow from `ghcid`'s source code, since if you offer a custom `--command`, all of those are lost (relative to what a vanilla `ghcid` call loads on a simple project). I don't like this approach because it isn't as scalable, and also is at risk of drifting from what `ghcid` intends with the flags it would otherwise load itself.

My question is this: can we extend `ghcid` to offer the following usage?
```
ghcid foo:lib foo:test:foo-tests
```
where otherwise all the defaults are loaded? From `ghcid --help` I see that it can natively accept a single Module to load, but here I'm looking to load an entire component while avoiding arcane invocations. 

Please and thank you. I'm willing to help with such a PR if pointed in the right direction.

fosskers:
This PR enables `--target` to be used with stack-based projects. It also allows `--target` to be passed more than once, since `stack` can load multiple components. 

Closes #297

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -104,7 +104,7 @@ options = cmdArgsMode $ Options
     ,color = Auto &= name "colour" &= name "color" &= opt Always &= typ "always/never/auto" &= help "Color output (defaults to when the terminal supports it)"
     ,setup = [] &= name "setup" &= typ "COMMAND" &= help "Setup commands to pass to ghci on stdin, usually :set <something>"
     ,allow_eval = False &= name "allow-eval" &= help "Execute REPL commands in comments"
-    ,target = Nothing &= typ "TARGET" &= help "Cabal target to build (e.g. lib:foo)"
+    ,target = Nothing &= typ "TARGET" &= help "Target Component to build (e.g. lib:foo for Cabal, foo:lib for Stack)"
     } &= verbosity &=
     program "ghcid" &= summary ("Auto reloading GHCi daemon v" ++ showVersion version)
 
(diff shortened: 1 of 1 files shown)


fosskers:
Perhaps this isn't 100%... I notice it loading everything twice when I first start up `ghcid`.

ndmitchell:
Thanks, looks good to me!

fosskers:
@ndmitchell do you know why I'd be seeing it load twice upon startup?

ndmitchell:
No idea! Are both loads into Ghci, or is one a compilation by Stack, followed by a load into Ghci? If it's two loads, then it may be stack loading the files, resetting something, and loading again. Does `stack ghci` in the same way load things twice?

fosskers:
Nope, only once.

ndmitchell:
No idea then, that does seem a bit odd. I'd welcome someone investigating, but probably not likely to do so myself. If you want, feel free to raise a separate bug report to so that we can accumulate our understanding of it.

ndmitchell:
I released 0.8.5 with this change included.

Issue #75, opened 2016-06-11
Colored output


talw:
It could be useful if ghcid had colored output.
Allows the eyes to parse those type check errors faster.

There is a solution for ghci, for example this script: https://github.com/rhysd/ghci-color.
However, this doesn't integrate readily with ghcid, at least not in a way I'm aware of.

Please excuse this issue if supporting such a feature is beyond the intended scope of ghcid, and perhaps the solution should be implemented externally. And if it is I'd appreciate a hint as to the form it should be implemented as.

ndmitchell:
It already has "coloured" output, it just so happens those two colours are normal and bold (aka light gray and bright white on most terminals). What were you thinking should be in different colors?

The relevant code would probably want to go near or in https://github.com/ndmitchell/ghcid/blob/master/src/Ghcid.hs#L226.

talw:
Oh of course, I did notice the bold parts, but I thought of something along the lines of:
https://wiki.haskell.org/wikiupload/a/ad/Coloured-error.png
But this depends on having the message parsed, or at least getting an AST output of something that parsed it.

ndmitchell:
I find that pretty ugly - highlighting instance and of despite not being keywords in this context, plus the general colour scheme seems to "bright". If we can come up with something that makes it easier to get at the relevant part of the error quickly that does seem useful though. I suspect that was done with just a lexer, not a full AST, so shouldn't be too hard to replicate if we do want to go that route.

talw:
You're right, I didn't refer to that exact coloring scheme, Just to the granularity level (not including marking non-keywords in different colors!)

Speaking of coloring scheme though, everyone has different backgrounds and color preferences, so I'm not so sure if there are sensible defaults we can use.
The approach you took with the bold was good in that regard, as it is neutral.

ndmitchell:
The question I'd like to consider first is what additional information would it helpful to highlight?

* File location? Perhaps, but it's always on the first line without indents, so in a way that's already clear.
* Brackets? Useful when matching brackets, but I rarely find that's the cause of an error in ghci (because you only really get the brackets out of the AST error messages, by which time you know they are correct)
* Literals? Again, these are correct by the time you see them in errors such as above.

talw:
The thing I thought would be helpful to highlight is basically type signatures.
Namely, the expected and actual type signatures, as they are the bread and butter when I'm fixing errors. I assume that's not just me.
[...]

PR #312, opened 2020-03-20
color output as suggested by #75

crvdgc:
Set file type to Haskell to enable syntax highlighting, as suggested by #75. Make this behavior default for vim.


ndmitchell:
Thanks! Looks totally plausible.

PR #314, opened 2020-04-08
Report GHC panics

basile-henry:
Here is an example of a GHC panic that `ghcid` was hiding:

```
ghc: panic! (the 'impossible' happened)
  (GHC version 8.6.5 for x86_64-unknown-linux):
        buildKindCoercion
  Div ((rp_a1iFz * (1 + (64 * (CLog 2 64 + 8)))) - 1) 18
  Div ((1 + (64 * (CLog 2 64 + 8))) - 1) WeightRamWidth
  CLog 2 64 + 8
  8
  Call stack:
      CallStack (from HasCallStack):
        callStackDoc, called at compiler/utils/Outputable.hs:1160:37 in ghc:Outputable
        pprPanic, called at compiler/types/Coercion.hs:2069:9 in ghc:Coercion

Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
```

This PR parses this type of error in order for `ghcid` to report them properly.

Note: the last line is not kept as part of the error message. Maybe it
should be, but it is not obvious to me how to do so since it doesn't
seem to follow the same leading whitespace format as the rest of the
error message.

Thanks for the pull request!

### DIFF
--- a/src/Language/Haskell/Ghcid/Parser.hs
+++ b/src/Language/Haskell/Ghcid/Parser.hs
@@ -62,11 +62,17 @@ parseLoad (map Esc -> xs) = nubOrd $ f xs
             | Just file <- stripPrefixE "<no location info>: can't find file: " x
             = Message Error (unescapeE file) (0,0) (0,0) [fromEsc x] : f xs
 
+        -- <no location info>: error:
+        f (x:xs)
+            | unescapeE x == "<no location info>: error:"
+            , (xs,rest) <- span leadingWhitespaceE xs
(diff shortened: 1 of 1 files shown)


basile-henry:
To give a bit more context, this was a GHC panic in a typechecker plugin.

I am not sure exactly what GHC panics end up with this type of error with no location/file information.
Thankfully I don't see GHC panics very often (even when `ghcid` is not there to hide them 😅).

ndmitchell:
Awesome work! A great idea.

ndmitchell:
Thanks, I've released 0.8.6 with this included.

PR #317, opened 2020-05-16
Activate on startup

chrismwendt:
I thought `"activationEvents":["workspaceContains"]` would trigger
activation whenever a matching file was added/changed/removed. It turns
out that workspaceContains is only evaluated once on startup:

https://code.visualstudio.com/api/references/activation-events#workspaceContains

cc @ndmitchell

### DIFF
--- a/plugins/vscode/package.json
+++ b/plugins/vscode/package.json
@@ -18,10 +18,7 @@
         "Programming Languages",
         "Linters"
     ],
-    "activationEvents": [
-        "onCommand:extension.startGhcid",
-        "workspaceContains:**/ghcid.txt"
-    ],
+    "activationEvents": ["*"],
     "main": "./out/src/extension",
     "contributes": {
         "commands": [
--- a/plugins/vscode/package.json
+++ b/plugins/vscode/package.json
@@ -18,6 +18,10 @@
         "Programming Languages",
         "Linters"
     ],
+    "//": [
+        "Activate unconditionally on startup to register file watchers.",
+        "See https://github.com/ndmitchell/ghcid/pull/317"
+    ],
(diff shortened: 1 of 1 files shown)


ndmitchell:
Does this mean the extension is _always_ activated? I thought VS Code really didn't like it as such extensions slow down the program for everyone.

chrismwendt:
Correct, this means the extension will always activate on VS Code startup.

To see how expensive it is, I checked the "Developer: Show Running Extensions" command, which shows 5ms for ghcid. For comparison, most of my other extensions take 10-100ms, with Git taking 324ms.

The ghcid extension's emitted JS code is ~20KB, which I'd call "very small".

If this seems too expensive, there's one more alternative I can think of: add another command like "extension.startWatcher" and tell users to run that command if their command-line ghcid began created `ghcid.txt` after starting VS Code.

ndmitchell:
Could that reasoning and a link to this ticket go into the JSON file at this point, so we have the logic and reasoning recorded for everyone who asks the same question. I'm OK with starting at startup, as you mention, it's tiny.

chrismwendt:
Yup, according to https://stackoverflow.com/a/14221781/2061958 npm will ignore a `"//"` key at the root of package.json. Updated.

ndmitchell:
Cool. I guess a new release of the extension is a good idea?

chrismwendt:
Yep

ndmitchell:
0.3.1 now out with that fix. Should show up at https://marketplace.visualstudio.com/items?itemName=ndmitchell.haskell-ghcid shortly.

chrismwendt:
🎉 Confirmed it's working on 0.3.1 👌

PR #319, opened 2020-06-06
Allow eval multiline commands

akrmn:
The idea here is to extend `--allow-eval` to allow comments like 

```haskell
{- $>
data Foo = Bar | Baz
  deriving (Generic, Eq, Ord, Show)
<$ -}
```

to be picked up as commands by ghcid. The effect should be the same as typing

```haskell
:{
data Foo = Bar | Baz
  deriving (Generic, Eq, Ord, Show)
:}
```

in a ghci session.

### DIFF
--- a/src/Session.hs
+++ b/src/Session.hs
@@ -149,7 +149,7 @@ performEvals ghci True reloaded = do
     fmap join $ forM cmds $ \(file, cmds') ->
         forM cmds' $ \(num, cmd) -> do
             ref <- newIORef []
-            execStream ghci (unwords $ lines cmd) $ \_ resp -> modifyIORef ref (resp :)
+            execStream ghci cmd $ \_ resp -> modifyIORef ref (resp :)
             resp <- unlines . reverse <$> readIORef ref
             pure $ Eval $ EvalResult file (num, 1) cmd resp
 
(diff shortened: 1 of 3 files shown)


src/Session.hs
```diff
@@ -164,7 +164,10 @@ splitCommands [] = []
 splitCommands ((num, line) : ls)
     | isCommand line =
           let (cmds, xs) = span (isCommand . snd) ls
-           in (num, unlines $ fmap (drop $ length commandPrefix) $ line : fmap snd cmds) : splitCommands xs
+           in (num, unwords $ fmap (drop $ length commandPrefix) $ line : fmap snd cmds) : splitCommands xs
+    | isMultilineCommandPrefix line =
+          let (cmds, xs) = break (isMultilineCommandSuffix . snd) ls
+           in (num, unlines (wrapGhciMultiline (fmap snd cmds))) : splitCommands (drop 1 xs)
```
ndmitchell, line 170: HLint suggests using `drop1` here.

src/Session.hs
```diff
@@ -173,7 +176,20 @@ isCommand = isPrefixOf commandPrefix
...
+multilineCommandPrefix = "{- $>"
+
+isMultilineCommandSuffix :: String -> Bool
+isMultilineCommandSuffix = (==) multilineCommandSuffix
+
+multilineCommandSuffix :: String
+multilineCommandSuffix = "<$ -}"
+
+wrapGhciMultiline :: [String] -> [String]
+wrapGhciMultiline xs = [":{"] <> xs <> [":}"]
```
ndmitchell, line 192: Can we use `++` for concat as its a bit simpler.

src/Session.hs
akrmn: thanks!

src/Session.hs
akrmn: sure! bit of a habit from overloaded lists

ndmitchell:
Thanks, looks good. I've released 0.8.7 with this patch.

PR #343, opened 2021-06-10
Add target to status output

mrtnpaolo:
Howdy, this patch adds the target to the status output.

This is especially useful when one has multiple `ghcid` instances running (e.g. working at the same time on an executable and libraries it depends on.)

I put in a couple extra headings for the errors and warnings to give the target a reasonable place to appear in and I matched GHCi's color scheme for those.

One thing I'm not sure about is what the window title format should be: maybe it's good enough as it is since it shows the project? I also do not have a windows machine to try and get a feel.

Would appreciate feedback!
Thanks.

### DIFF
--- a/src/Ghcid.hs
+++ b/src/Ghcid.hs
@@ -290,7 +290,7 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
         outputFill currTime load evals msg = do
             load <- pure $ case load of
                 Nothing -> []
-                Just (loadedCount, msgs) -> prettyOutput currTime loadedCount (filter isMessage msgs) evals
+                Just (loadedCount, msgs) -> prettyOutput opts currTime loadedCount (filter isMessage msgs) evals
             TermSize{..} <- termSize
             let wrap = concatMap (wordWrapE termWidth (termWidth `div` 5) . Esc)
             (msg, load, pad) <-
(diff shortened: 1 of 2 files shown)


src/Language/Haskell/Ghcid/Util.hs
```diff
@@ -85,8 +85,16 @@ ignored act = do
     waitBarrier bar
 
 -- | The message to show when no errors have been reported
-allGoodMessage :: String
-allGoodMessage = setSGRCode [SetColor Foreground Dull Green] ++  "All good" ++ setSGRCode []
+allGoodMessage :: [String] -> String
+allGoodMessage target = setSGRCode [SetColor Foreground Dull Green] ++ "All good " ++ setSGRCode [] ++ "in " ++ concat target
```
ndmitchell, line 89: If target is blank, as it often will be, this will look really odd - i suspect you need to hide the "in " in that case.

src/Ghcid.hs
```diff
@@ -433,11 +433,16 @@ runGhcid session waiter termSize termOutput opts@Options{..} = do
...
 -- | Given an available height, and a set of messages to display, show them as best you can.
-prettyOutput :: String -> Int -> [Load] -> [EvalResult] -> [String]
-prettyOutput currTime loadedCount [] evals =
-    (allGoodMessage ++ " (" ++ show loadedCount ++ " module" ++ ['s' | loadedCount /= 1] ++ ", at " ++ currTime ++ ")")
+prettyOutput :: Options -> String -> Int -> [Load] -> [EvalResult] -> [String]
+prettyOutput Options{..} currTime loadedCount [] evals =
+    (allGoodMessage target ++ " (" ++ show loadedCount ++ " module" ++ ['s' | loadedCount /= 1] ++ ", at " ++ currTime ++ ")")
         : concatMap printEval evals
-prettyOutput _ _ xs evals = concatMap loadMessage xs ++ concatMap printEval evals
+prettyOutput Options{..} _ _ xs evals =
```
ndmitchell, line 440: My concern is that if you have only a handful of lines (e.g. 5) then in the case where you have errors/warnings, you've added 1 header line and decreased the amount of error/warning to display - potentially by a meaningful amount.

ndmitchell (commented):
Adding the targets to the all good message and the status bar seems reasonable, I'd be happy with that using a few cleanups. Less convinced about warnings/errors, since that is more precious real estate.

I wonder if the targets are blank if the basename of the current directory might be useful to include instead?

PR #357, opened 2022-03-15
vim: add split direction option

goolord:
i just like having the ghcid window open to the right, but using a autocmd to do some wincmds has been fussy for me :9


ndmitchell:
Thanks!

PR #366, opened 2022-09-19
fixing build issue with fsnotify

vtuizat:
I couldn't build ghcid with the latest versioin of fsnotify (v0.4).

This adds a constraint on the version of fsnotify to get the older version and be able to build ghcid.


ndmitchell:
I released a version yesterday that should work with 0.4. Does that not work for you? I've added a bound in the test suite to require the latest version too - so if not, pull again and try.

PR #370, opened 2022-12-06
more file watch utilities for FAQ

peterbecich:
Some more suggestions for the FAQ, copied from https://github.com/NorfairKing/feedback#comparison-with-other-tools


ndmitchell:
Thanks!

PR #373, opened 2023-03-06
Infer ranges

chrismwendt:
Some GHC error messages have range info:

```
foo/Bar.hs:23:4-6: error: ...
```

And some don't, but they have annotations:

```
foo/Bar.hs:23:4: error:
...
   |
57 |         run source srv
   |             ^^^^^^
```

This PR parses the `^^^` annotations to infer range info so that the red squiggles show up on the whole range in VS Code, not just the first character.

In the process, I had to do some housekeeping: update TypeScript, mocha, migrate to `@vscode/test-electron`, etc.

cc @ndmitchell

### DIFF
--- a/plugins/vscode/.gitignore
+++ b/plugins/vscode/.gitignore
@@ -1,2 +1,3 @@
 out
 node_modules
+.vscode-test
--- a/plugins/vscode/.gitignore
+++ b/plugins/vscode/.gitignore
@@ -1,3 +1,4 @@
 out
 node_modules
 .vscode-test
+yarn-error.log
--- a/plugins/vscode/package.json
+++ b/plugins/vscode/package.json
@@ -54,7 +54,7 @@
         "@types/mocha": "^2.2.48",
         "@types/node": "^6.14.9",
         "mocha": "^5.2.0",
-        "typescript": "^3.8.3",
+        "typescript": "^4.9.5",
         "vsce": "^1.74.0",
         "vscode": "^1.1.26"
     }
--- a/plugins/vscode/package.json
+++ b/plugins/vscode/package.json
@@ -47,15 +47,16 @@
     "scripts": {
         "vscode:prepublish": "tsc -p ./",
         "compile": "tsc -watch -p ./",
-        "postinstall": "node ./node_modules/vscode/bin/install",
-        "test": "node ./node_modules/vscode/bin/test"
+        "test": "node out/test/runTest.js"
     },
     "devDependencies": {
-        "@types/mocha": "^2.2.48",
+        "@types/glob": "^8.1.0",
+        "@types/mocha": "^10.0.1",
         "@types/node": "^6.14.9",
-        "mocha": "^5.2.0",
+        "@types/vscode": "^1.76.0",
+        "@vscode/test-electron": "^2.3.0",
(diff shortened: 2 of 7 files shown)


chrismwendt:
@ndmitchell It looks like the copyright year is out of date in CI:

https://github.com/ndmitchell/ghcid/actions/runs/4345680401/jobs/7590729342?pr=373#step:7:105

ndmitchell:
Thanks for the patch. Should GHC HQ be adding the range if they clearly know it? It seems a bit weird to parse it out of the string. Have you reported this upstream?

I actually don't use the IDE extension anymore. So if you want to take over maintainership of this extension, you'd be most welcome to do so.

ndmitchell:
Also, I wonder if this range inference should be in the Haskell side, if we do want to have it?

chrismwendt:
Including the range info would be convenient for parsing, but on the other hand redundant with the annotation. I haven't reported this upstream since parsing makes it work on existing GHC versions.

I've been using HLS more recently and I don't use the ghcid IDE features much anymore either, so I don't think I'd be a responsive maintainer.

> Also, I wonder if this range inference should be in the Haskell side, if we do want to have it?

Do you mean ghcid would read the GHC output, parse it, infer ranges, append the range info to errors, then print the augmented errors? I suppose the other plugins (nvim, emacs) would benefit from the info, too.

chrismwendt:
Found a repro. I'm on 57831ecca9d2e2359391f6aa8adde5ed952676e3 with this diff:

```diff
diff --git a/src/Session.hs b/src/Session.hs
index 343bd9f..5f4bb26 100755
--- a/src/Session.hs
+++ b/src/Session.hs
@@ -74,7 +74,7 @@ kill ghci = ignored $ do
         ignored $ quit ghci
         debugShutdown "After quit"
     debugShutdown "Before terminateProcess"
-    ignored $ terminateProcess $ process ghci
+    ignored $ terminateProcessBLAH $ process ghci
     debugShutdown "After terminateProcess"
     -- Ctrl-C after a tests keeps the cursor hidden,
     -- `setSGR []`didn't seem to be enough
```

Run this command:

```
ghcid --command "cabal repl exe:ghcid" --output ghcid.txt
```

Notice `src/Session.hs:77:15:` is missing the column range in `ghcid.txt`:

```
src/Session.hs:77:15: error: [GHC-88464]
    Variable not in scope:
      terminateProcessBLAH :: ProcessHandle -> IO ()
    Suggested fix:
      Perhaps use ‘terminateProcess’ (imported from System.Process)
   |
77 |     ignored $ terminateProcessBLAH $ process ghci
   |               ^^^^^^^^^^^^^^^^^^^^
```

Side note: if you change it to `terminateProcessBLAH2` to reload, then `ghcid.txt` has the range (seems to be an issue on initial load, not subsequent reloads):

```
src/Session.hs:77:15-35: error: [GHC-88464]
    Variable not in scope:
      terminateProcessBLAH2 :: ProcessHandle -> IO ()
    Suggested fix:
      Perhaps use ‘terminateProcess’ (imported from System.Process)
   |
77 |     ignored $ terminateProcessBLAH2 $ process ghci
   |               ^^^^^^^^^^^^^^^^^^^^^
```

PR #375, opened 2023-06-01
Fix crash when modules are renamed or deleted / `withFile: does not exist`

9999years:
With GHC 9.6, renaming or deleting a module while `ghcid` is running will trigger a crash with a confusing error message:

```
Reloading...
  ~/repo/src/MyModule.hs

All good (0 modules, at 12:40:11)

No files loaded, nothing to wait for. Fix the last error and restart.
```

The `unescapeE x == "<no location info>: error:"` check doesn't work with GHC 9.6 because errors have unique numbers attached now:

```
<no location info>: error: [GHC-82272]
    module 'MyCoolModule' cannot be found locally
```

After fixing the message parsing logic, we get another confusing error message:

```
All good (5669 modules, at 14:21:52)
Reloading...
  ~/repo/src/MyCoolModule.hsghcid: ~/repo/<unknown>: withFile: does not exist (No such file or directory)
```

It turns out that `Message`s with `loadFile = "<unknown>"` aren't properly filtered in `loadedModules`:

https://github.com/ndmitchell/ghcid/blob/ac3c2f922f3cd7157f58ee730010e4d478cb3499/src/Session.hs#L85

This PR fixes both issues.

### DIFF
--- a/src/Language/Haskell/Ghcid/Parser.hs
+++ b/src/Language/Haskell/Ghcid/Parser.hs
@@ -64,7 +64,7 @@ parseLoad (map Esc -> xs) = nubOrd $ f xs
 
         -- <no location info>: error:
         f (x:xs)
-            | unescapeE x == "<no location info>: error:"
+            | "<no location info>: error:" `isPrefixOfE` x
             , (xs,rest) <- span leadingWhitespaceE xs
             = Message Error "<unknown>" (0,0) (0,0) (map fromEsc $ x:xs) : f rest
 
--- a/src/Session.hs
+++ b/src/Session.hs
@@ -81,11 +81,15 @@ kill ghci = ignored $ do
     -- See: https://github.com/ndmitchell/ghcid/issues/254
     showCursor
 
-loadedModules :: [Load] -> [FilePath]
-loadedModules = nubOrd . map loadFile . filter (not . isLoadConfig)
+loadedModules :: FilePath -> [Load] -> [FilePath]
+loadedModules dir = nubOrd . map (loadFile . qualify dir) . filter predicate
+    where
+      predicate Message{loadFile = loadFile} = loadFile /= "<unknown>"
+      predicate Loading{loadFile = loadFile} = loadFile /= "<unknown>"
+      predicate _ = False
 
-qualify :: FilePath -> [Load] -> [Load]
-qualify dir xs = [x{loadFile = dir </> loadFile x} | x <- xs]
+qualify :: FilePath -> Load -> Load
+qualify dir message = message{loadFile = dir </> loadFile message}
 
 -- | Spawn a new Ghci process at a given command line. Returns the load messages, plus
 --   the list of files that were observed (both those loaded and those that failed to load).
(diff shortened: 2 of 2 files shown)


src/Session.hs
```diff
@@ -81,11 +81,15 @@ kill ghci = ignored $ do
     -- See: https://github.com/ndmitchell/ghcid/issues/254
     showCursor
 
-loadedModules :: [Load] -> [FilePath]
-loadedModules = nubOrd . map loadFile . filter (not . isLoadConfig)
+loadedModules :: FilePath -> [Load] -> [FilePath]
+loadedModules dir = nubOrd . map (loadFile . qualify dir) . filter predicate
+    where
+      predicate Message{loadFile = loadFile} = loadFile /= "<unknown>"
+      predicate Loading{loadFile = loadFile} = loadFile /= "<unknown>"
```
9999years, line 88: Is there a nice way to combine these cases?

9999years:
There might be a similar issue with the error parsing here:

https://github.com/ndmitchell/ghcid/blob/e2852979aa644c8fed92d46ab529d2c6c1c62b59/src/Language/Haskell/Ghcid/Parser.hs#L74

src/Language/Haskell/Ghcid/Parser.hs
```diff
@@ -64,7 +64,7 @@ parseLoad (map Esc -> xs) = nubOrd $ f xs
 
         -- <no location info>: error:
         f (x:xs)
-            | unescapeE x == "<no location info>: error:"
+            | "<no location info>: error:" `isPrefixOfE` x
```
evanrelf, line 67: Should we still remove escape characters here?

```suggestion
            | "<no location info>: error:" `isPrefixOfE` unescapeE x
```

src/Session.hs
```diff
@@ -81,11 +81,15 @@ kill ghci = ignored $ do
     -- See: https://github.com/ndmitchell/ghcid/issues/254
     showCursor
 
-loadedModules :: [Load] -> [FilePath]
-loadedModules = nubOrd . map loadFile . filter (not . isLoadConfig)
```
evanrelf, line 85: This was the only use of `isLoadConfig`, so you could remove it from `Language.Haskell.Ghcid.Types` ([link](https://github.com/ndmitchell/ghcid/blob/e2852979aa644c8fed92d46ab529d2c6c1c62b59/src/Language/Haskell/Ghcid/Types.hs#L74C1-L77)).

There is no `isEval` function to go with the `Eval` data constructor, so you wouldn't be disrupting a complete set of functions, if you will.

src/Session.hs
evanrelf: I don't think so.

evanrelf (approved):

src/Language/Haskell/Ghcid/Parser.hs
9999years: Nah, `unescapeE` is an `Esc -> String` transform, `isPrefixOfE` handles the `Esc` escaping already.

src/Session.hs
parsonsmatt: Or patterns will help here

src/Session.hs
ndmitchell: The fact that `isLoadConfig` isn't used elsewhere isn't a big deal - it's still an exported function on a public type, so has general utility.

ndmitchell:
Thanks for the fix and the review on it!

PR #376, opened 2023-06-02
Add `--start-command`

9999years:


src/Ghcid.hs
```diff
@@ -162,15 +164,21 @@ autoOptions o@Options{..}
...
+              | otherwise -> updateOptions ("ghci":opts) []
     where
-        f c r = o{command = unwords $ c ++ map escape arguments, arguments = [], restart = restart ++ r, run = [], test = run ++ test}
+        updateOptions command' restart' =
+          o { command = unwords $ command' ++ map escape arguments
+            , arguments = []
+            , restart = restart ++ restart'
+            , run = []
+            , test = run ++ test
+            }
```
9999years, line 181: Drive-by fix to give these variables better names + format them a little more clearly

9999years:
Somehow, ghcid is architected in a way that makes detecting new files and directories extremely painful. I've decided it'll be easier to reimplement ghcid than fix it.

PR #378, opened 2023-06-28
Write output of the linter into output file(s)

KAction:
Output file enables integration with editors to parse compiler error messages and jump to the error location, and there is no reason to not have this convenience with linter.

By the way, thank you for the project. It is awesome. It provides one of the best developer experiences across all tools and all languages I worked with.


ndmitchell:
Thanks for the patch and the kind remarks, looks like a good idea!

PR #418, opened 2026-07-28
Fix --server

tomjaguarpaw:
https://github.com/ndmitchell/ghcid/commit/daf89d0ca7430f2abbc2c9175ab1e89ab1fc9926 broke the test suite, and probably the executable. I guess something like this was intended instead.


ndmitchell:
Thanks!

PR #425, opened 2026-09-14
Fix change detection with '-fobject-code'

alexfmpe:
When '-fobject-code' is enabled, ghci will cache compilation across runs so we only see "[X of Y] Compiling ..." messages for modules that were modified (or had their deps be) since the last run. When all (or none) modules are cached, ghcid works as usual, but when only some are, only the modules triggering messages are watched. Once one of those is modified, then 'sessionReload' kicks in and all ':show modules' are watched again, so we adopt that source of file paths in 'sessionStart'.


app/Session.hs
```diff
@@ -125,7 +125,11 @@ sessionStart Session{..} cmd setup = do
     writeIORef pathMode $ determinePathMode moduleFiles
     messages <- pure $ map (qualify dir) messages
 
-    let loaded = loadedModules dir messages
+    loaded <- do
+        Just ghci <- readIORef ghci
+        shownModules <- showModules ghci
+        pure $ nubOrd $ loadedModules dir messages <> map ((dir </>) . snd) shownModules
```
alexfmpe, line 131: I originally simply replaced `loadedModules` with `shownModules` as the source but that broke the `Cd basic` test.

